PR #53, opened 2025-06-12
CMake pt.2: Electric Boogaloo 

DolphinChips:


uvcat7:
@sukibaby, are you still having issues with this? Looks like it's working for me.

DolphinChips:
Now it's fully migrated to vcpkg&CMake, including CI \o/ Another nice thing to have is binary cache for CI, but I don't have enough permissions for that, instructions for it can be found at https://learn.microsoft.com/en-us/vcpkg/users/binarycaching

.github/workflows/codeql.yml
```diff
@@ -30,7 +30,12 @@ jobs:
...
         - language: c-cpp
-          build-mode: autobuild
+          build-mode: manual
+          configure: >
+            cmake -Bbuild
+            -DCMAKE_TOOLCHAIN_FILE="$env:VCPKG_INSTALLATION_ROOT\scripts\buildsystems\vcpkg.cmake"
+            -DVCPKG_TARGET_TRIPLET=x64-windows-static-md
+          build: cmake --build build --parallel "$env:NUMBER_OF_PROCESSORS"
         - language: python
           build-mode: none
```
ScottBrenner:
```suggestion
```
https://github.com/uvcat7/ArrowVortex/actions/runs/16276299442/job/45955901387#step:7:63

.github/workflows/windows.yml
```diff
@@ -22,8 +22,16 @@ jobs:
...
+        run: >
+          cmake -Bbuild -DCMAKE_BUILD_TYPE=Release
+          -DCMAKE_TOOLCHAIN_FILE="$env:VCPKG_INSTALLATION_ROOT\scripts\buildsystems\vcpkg.cmake"
+          -DVCPKG_TARGET_TRIPLET=x64-windows-static-md
       - name: Build AV
-        run: msbuild build\VisualStudio\ArrowVortex.vcxproj /p:Configuration=Release /p:Platform=x64         
+        run: cmake --build build --config Release --parallel "$env:NUMBER_OF_PROCESSORS"
+      - name: Collect into a directory
+        if: github.ref_name == 'beta'
+        run: cmake --install build --config Release --prefix ./
```
ScottBrenner, line 52:
```suggestion
```
If the previous step builds into the `build/` directory it can be referenced in the upload step below without a separate step to collect things

DolphinChips:
FWIW currently there's a regression from File.cpp changes which revealed a bug in how executable is launched, so this shouldn't be merged for now

.github/workflows/windows.yml
```diff
@@ -22,8 +27,29 @@ jobs:
...
       - name: Upload artifact
         if: github.ref_name == 'beta'
         uses: actions/upload-artifact@v4
```
ScottBrenner, line 55: `path` below should be modified to `build/src/System/Release/ArrowVortex.exe`, per https://github.com/uvcat7/ArrowVortex/pull/53/checks#step:7:156

DolphinChips: `cmake --install` copies it and assets to `$INSTALL_PATH/bin/`, in this case just `./bin`, see https://github.com/uvcat7/ArrowVortex/actions/runs/16259761839

uvcat7:
Reminder to myself that C++ 20 needs to be enabled before merging but after rebase.

nobbele:
> FWIW currently there's a regression from File.cpp changes which revealed a bug in how executable is launched, so this shouldn't be merged for now

What's the issue?

DolphinChips:
> > FWIW currently there's a regression from File.cpp changes which revealed a bug in how executable is launched, so this shouldn't be merged for now
> 
> What's the issue?

#144

convextriangle:
> > > FWIW currently there's a regression from File.cpp changes which revealed a bug in how executable is launched, so this shouldn't be merged for now
> > 
> > 
> > What's the issue?
> 
> #144

it should have been resolved in #145 - are there any other blockers for merging this?

convextriangle:
merge mistake, will need to be force pushed over likely
my bad

.github/workflows/codeql.yml
```diff
@@ -48,6 +56,29 @@ jobs:
...
         build-mode: ${{ matrix.build-mode }}
 
+    - name: Add NuGet sources
+      shell: pwsh
+      if: matrix.language == 'c-cpp'
+      run: |
+        .$(vcpkg fetch nuget) `
+          sources add `
+          -Source "${{ env.FEED_URL }}" `
+          -StorePasswordInClearText `
```
ScottBrenner, line 66: What does this flag do?

DolphinChips: Stores password in clear text :P According to documentation,
> Be aware that encrypted passwords are only supported on Windows.

Binary caching was setup according to https://learn.microsoft.com/en-us/vcpkg/consume/binary-caching-github-packages?pivots=windows-runner, without double-checking whatever it does

uvcat7 (approved):
Okay, I have NO idea what you changed to solve the problem but this finally seems to work correctly from a fresh copy of the source code.

The build information to fix the freetype issue should be good enough if this happens in the future, but I really hope it doesn't.

PR #71, opened 2025-07-06
Label Manager

Psycast:
label_manager

Implements a dedicated Label/Bookmark manager for quick jumping around a file and makes labels generally useful.

**This also makes Labels get saved into a .sm along with .ssc, this is outside spec but no issues have appeared due to this change.**

Also add prevention for characters `;,=` that cause file loading issues in AV and games.

### Includes:
- Clicking a list item will jump to that section of the song.

- A unit switch that changes the displayed timestamp as either:
   - Time
   - Beat
   - Row

- A copy to clipboard button for quick sharing for things such as notes between users. This uses the current time unit selected.


Psycast:
Is there any specific reason you are against saving them in .sm?

afaict, saving non-standard data to the files has become more common such as GrooveAuthor saving editor specific data for itself.

uvcat7:
@teejusb, you're the Stepmania expert--does this cause an issue or is it irrelevant?

teejusb:
Even if it doesn't cause an issue, I agree with @sukibaby.

I also don't think we should be saving extra fields if the format doesn't make use of them. It makes the files themselves non-standard and really just causes confusion for people who wouldn't know otherwise. Just because GA does it, which already sounds non-ideal, doesn't mean AV should as well. If people want editor specific data, that should probably a supplemental file and not jammed into the sm imo.

Gamewise, it's not useful unless the theme somehow makes use of it, which it shouldn't since that's not its job to do so (the engine should generally take care of parsing the sm/sscs). I also think `.sm` is essentially legacy and pushing for `.ssc` is the way to go. The `.sm` spec is basically frozen since `.ssc` already supports everything and more.

uvcat7:
Thanks for the input!
@Psycast if you wouldn't mind removing the label saves from .sm, I will test and merge.

Psycast:
Might be worth adding a warning to saving a .sm if any SM5 features are used that they won't be saved.

teejusb:
Here are the tags that get parsed by `.sm` files:
https://github.com/itgmania/itgmania/blob/release/src/NotesLoaderSM.cpp#L202-L236

>Might be worth adding a warning to saving a .sm if any SM5 features are used that they won't be saved.

Agreed! I think it's good to give users more insight into their files so they learn what is correct/incorrect.

src/Dialogs/LabelBreakdown.cpp
```diff
@@ -0,0 +1,382 @@
...
+	String out;
+	for (auto button : myButtons)
+	{
+		out += button->myDisplayTime;
+		out += ": ";
+		out += button->myDisplayText;
+		out += "\n";
+	}
+	if(out.empty()) {
+		HudInfo("%s", "There is no labels to copy...");
```
uvcat7: *are

nobbele:
Why not put these tags in comments?
Having editor metadata in comments sounds reasonable to me

StarbotArc:
That's actually a pretty effective idea. Honestly if it works on most simulators then it'd be the best solution.

Psycast:
This was generally to expand the usefulness of labels already preset in ssc files, not to add AV specific metadata.

nobbele:
I see, possible future feature then

uvcat7 (approved):
looks great!

PR #79, opened 2025-07-07
Seperate Chart Previewer + XMOD Support

### LINKED ISSUE #22: Add .ssc scroll/gimmick previews to chart preview

Psycast:
This integrates a separate preview notefield with complete support for scroll/Speed/Fake segments.

The decision to use a separate notefield is due to these reasons:
- Reduces the amount of changes required to handle the non-uniform positioning for things like tempoboxes + tooltips, selections, etc.
- Allows the preview to better mimic gameplay, as Etterna limits the amount of draw ahead to 20 beats / 960 rows, which some files use for effect.
- Fake Notes and Fake regions are transparent to inform the mapper.
- Editing can see the changes in realtime without it being difficult to edit, as using 0 scroll/speed collapses everything onto itself.
- Preview field can be moved side to side.
- Can change the preview to cmod/xmod/all, separate from editor view.
- Can toggle reverse scroll, separate from editor view. This is also useful for recording videos with both scroll directions at the same time.

Future things for better user experience:
- Remember window size / state.

Closes #22, #7


uvcat7 (approved):
Looks great! This is super exciting; thanks for the hard work.

PR #86, opened 2025-07-09
Insert Beats Changes

Psycast:
- Don't move tempo objects at row 0 to avoid multiple issues.
- Adjust tempo objects on a single chart when only a sim tempo is found.
- Set cursor to start of offset when undoing/redoing.


uvcat7 (approved):
Works good

PR #106, opened 2025-07-10
Drag-n-drop loading improvements

sukibaby:
Since drag-n-drop loading seems to still be in a fragile state I made some new improvements for file loading.

First, the drag-n-drop loading logic is taken out of the `handleMsg` switch case and moved into its own function `handleFileDrop`.

The function is then expanded with extra safeties to try and prevent crashes due to a race condition, or crashes due to really long paths to the files.

Btw,  MP3 is more prone to crashing when randomly dragging and dropping files than OGG is. This PR makes MP3 loading a lot less prone to crashing the app, but this doesn't completely 100% resolve weird behavior with specifically loading MP3 this way. I haven't looked into the LoadMp3 code yet, so I'll check it later to see if there's something we can be doing better.

### DIFF
--- a/src/System/System.cpp
+++ b/src/System/System.cpp
@@ -36,6 +36,7 @@
 #include <bitset>
 #include <list>
 #include <vector>
+#include <exception>
 
 #undef DELETE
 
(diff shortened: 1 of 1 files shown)


uvcat7:
I think my commit to #108 fixes this. Can you test/confirm?

sukibaby:
> I think my commit to #108 fixes this. Can you test/confirm?

It does seem to, so I'll close this and you can merge 108 🙏

PR #107, opened 2025-07-10
Replace all typedefs with standard types

sukibaby:
Removes the following typedefs from Core.h, so that only standard types are used:

typedef uint64_t  ulong
typedef uint32_t   uint
typedef uint16_t ushort
typedef uint8_t  uchar
typedef uint32_t   color32

Also the Windows macro UINT is used in the code base so it just makes things a  bit cleaner to explicitly use uint32_t instead of lowercase uint.

Note I don't find much value in the color32 typedef (I prefer knowing it's a 32 bit unsigned), but if others prefer it being color32, I don't feel super strongly about it.

### DIFF
--- a/src/Core/ByteStream.cpp
+++ b/src/Core/ByteStream.cpp
@@ -13,7 +13,7 @@ namespace Vortex {
 
 WriteStream::WriteStream()
 {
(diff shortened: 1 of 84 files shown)


uvcat7 (approved):
LGTM

PR #108, opened 2025-07-10
Improve stability of MP3 loading

sukibaby:
Switching to unique_ptr / make_unique manages the lifetime of the object automatically and also handles cleanup automatically. Addresses a common crash when loading MP3 via drag and drop.

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -1,8 +1,12 @@
-name: Windows Build
-
 on:
   push:
-    branches: ["main"]
+  pull_request:
+    branches:
+      - beta
+    paths-ignore:
+      - '**.md'
+
+name: Latest Beta
 
 jobs:
   windows:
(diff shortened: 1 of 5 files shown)


uvcat7:
This still crashes if you drag multiple MP3 files into the editor at once.

uvcat7:
I fixed the crash that was happening on my version. It happens more often with higher FPS (so turn your vsync off to test), and won't happen at all with multithreading disabled.
I tried like 20-30 copies and no crashes now.

sukibaby:
It seems to work just as well here, so I'll approve it.

Fwiw regarding the FPS thing, I typically use AV on a 60hz monitor, but I have a 165 and a 180hz monitor as well. In the future I'll be sure to test changes on both the higher refresh rate monitors.

uvcat7:
FWIW this seems to have been most common on mp3 because the editor converts to .ogg internally which takes a second

PR #112, opened 2025-07-11
Fix wav format compliance

sukibaby:
The proposed implementation is more correct, and resolves bugs where certain types of valid wav files (such as 16 bit 48khz) would be considered invalid by AV.


uvcat7 (approved):
Yeah LGTM, I don't use .wav ever

PR #121, opened 2025-07-11
Refactor pathing in workflow

ScottBrenner:
The workflow currently works in and creates additional directories, this proposes refactoring with default paths to clean things up a bit. Also refactors the 'Collect into a directory' step a bit, notably removing usage of `7z` as actions/upload-artifact compresses the path by default https://github.com/actions/upload-artifact?tab=readme-ov-file#altering-compressions-level-speed-v-size

Example run in my fork - https://github.com/ScottBrenner/ArrowVortex/actions/runs/16231746598

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -10,40 +10,31 @@ jobs:
   windows:
     name: "Build Windows x64"
     runs-on: windows-2022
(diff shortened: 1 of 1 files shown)


sukibaby (approved):
lgtm

PR #129, opened 2025-07-13
Replace with FileReader/Writer with std::fstream

nobbele:
Upgrades to C++17 because of \<filesystem\> (simplest way to get a `file_size` function), but there are no error-level compatibility problems.

Messy merge with beta and linux-4-mingw32 :(


uvcat7:
Thanks, this looks great for the most part. Can you just pull out the fstream source code files and throw them on a new branch forked from beta?
We'll be using CMake, not mingw, for builds going forward so I don't want to use those.

StarbotArc:
Should it be ideally just source, so we don't have workflow conflicts?

uvcat7:
Yes, exactly.

nobbele:
> Thanks, this looks great for the most part. Can you just pull out the fstream source code files and throw them on a new branch forked from beta? We'll be using CMake, not mingw, for builds going forward so I don't want to use those.

yeah of course that's why it's just a draft

also wdym "We'll be using CMake, not mingw"? the mingw build is using cmake, or are you planning to just not include the cross-compilation support?

StarbotArc:
I don't think it should be implemented juuuuust yet, I still need to tweak it and make it work out. I'll make a PR when I do have the Linux-CMake branch (linux-4-mingw32) done, as I usually do.

uvcat7:
> also wdym "We'll be using CMake, not mingw"? the mingw build is using cmake, or are you planning to just not include the cross-compilation support?

Sorry, I mean there's no problem to include cross-compilation so you two can develop, but I don't want to officially support it until we can truly make AV multiplatform.

StarbotArc:
(this discussion is probably going beyond what this pr is about, we should probably discuss else where)

Issue #92, opened 2025-07-09
Replace String classes with std::string


uvcat7:

uvcat7:
WString still needs to be addressed before this issue can be closed.

PR #130, opened 2025-07-13
Replace String/StringRef with std::string

poco0317:
This mostly handles #92 but does not replace WideString(.h/.cpp). Because of that, I don't think I would say this closes that issue. wstring and related things are just disgusting to me so I hope this is good enough for now. At that point you also have to consider its behavior on multiple platforms.

This was done by first essentially find-replacing all instances of `String` with `std::string` and `StringRef` with `const std::string&`. Then I went through and made it compile. It worked almost perfectly immediately after.
One might consider to replace `const std::string&` with `std::string_view` but I have not enough experience with that to mess with it. They are basically the same thing.

Doing this also required rewriting a several bits of code or references:
- `String.str()` was replaced almost entirely by `std::string.c_str()`
- `String.begin()` was replaced almost entirely by `&std::string[0]`
- `String._string` was replaced almost entirely by `std::string.data()`
- Functions in StringUtils.cpp are replaced with STL code or just modified a bit so they compile and run. I didn't entirely remove the utility functions in StringUtils because STL does not replace them 1 to 1. A good example of this is `IsWhiteSpace(c)` vs `std::isspace(c)` which return very similar results but `std::isspace` has more coverage. It's not part of my intention here to add functionality directly by doing that. So functions that refer to white space still point at `IsWhiteSpace` in such a way that you could either replace the content of `IsWhiteSpace` with `std::isspace` or replace it at a higher level. Anyways...
- Added an `#include <Core/Core.h>` to a few files. It wouldn't compile without it. Didn't ask why. Also added `#include <string>` and `#include <algorithm>` where applicable.
- `String operator +=` was replaced by `x = x + y` inline and all instances of `y` being a char are replaced by a string.
- Did not comprehensively search for StringUtils functions which are now obsoleted or can be replaced collectively by a single STL call

I might suggest adding [fmt](https://github.com/fmtlib/fmt) to replace Fmt in StringUtils. Thoughts for another time. 

I tested this by compiling AV and then opening and saving a few files. I also reopened AV and made sure the settings load and save properly. The files I opened contained both regular characters in the path and .sm and also irregular characters (mojibake and japanese characters). For example, files from BangerZ and mizuki were the ones picked. Definitely needs a bit of attention to testing because this affects basically every part of the application.


StarbotArc:
Thank you, Poco. Though I will say that `std::format` exists, we are in talks of *maybe* moving to C++20, but that's for a later time.

StarbotArc:
Alright, reviewed and merged. Had to touch some stuff up, but should be good to go.

DolphinChips:
Re: fmt, i'm already using it as a part of Vector migration

src/Core/Shader.cpp
```diff
@@ -149,16 +149,17 @@ static GLuint Compile(int type, const char* code, String& log)
 		{
 			int size = 1;
 			glGetShaderiv(shader, GL_INFO_LOG_LENGTH, &size);
-			log = String(size, 0);
-			glGetShaderInfoLog(shader, size, nullptr, log.begin());
+			log = "";
```
nobbele: `log.clear()`? same on line 200

src/Core/StringUtils.cpp
```diff
@@ -16,215 +17,143 @@ inline int max(int a, int b) { return (a < b) ? b : a; }
...
-	StrSetLen(s.string_, len - size);
-}
-
-}; // Str2
-
 // ================================================================================================
 // Str :: assign functions.
 
-String Str::create(const char* begin, const char* end)
+std::string Str::create(const char* begin, const char* end)
```
nobbele: `std::string(begin, end)` does the same thing

src/Core/StringUtils.cpp
```diff
@@ -16,215 +17,143 @@ inline int max(int a, int b) { return (a < b) ? b : a; }
...
 	else
 	{
 		s.clear();
 	}
 }
 
-void Str::assign(String& s, String&& str)
+void Str::assign(std::string& s, std::string&& str)
 {
 	s.swap(str);
```
nobbele: `s = str` should be more efficient and (hopefully do the same thing)

src/Core/StringUtils.cpp
```diff
@@ -16,215 +17,143 @@ inline int max(int a, int b) { return (a < b) ? b : a; }
...
 {
-	assign(s, str.string_, str.len());
+	assign(s, str.data(), static_cast<int>(str.length()));
 }
 
-void Str::assign(String& s, const char* str)
+void Str::assign(std::string& s, const char* str)
 {
-	assign(s, str, strlen(str));
+	assign(s, str, static_cast<int>(strlen(str)));
```
nobbele: these could just call the copy and C string constructors respectively instead of these unnecessary overload calls?

src/Core/StringUtils.cpp
```diff
@@ -5,6 +5,7 @@
 #include <stdio.h>
 #include <stdint.h>
 #include <math.h>
+#include <algorithm>
 
 # pragma warning(disable : 4996) // stricmp.
```
nobbele: can be removed, seems like stricmp has been replaced

src/Simfile/SaveSm.cpp
```diff
@@ -42,18 +42,18 @@ struct ExportData
...
-			"which might not work on some versions of Stepmania/ITG.", name.str());
+			"which might not work on some versions of Stepmania/ITG.", name.c_str());
 	}
 }
 
-static String Escape(const char* name, const char* str)
+static std::string Escape(const char* name, const char* str)
 {
-	String s = (String)str;
+	std::string s = (std::string)str;
```
nobbele: use constructor instead?

nobbele (commented):
I didn't test the PR but the code looks good to me
I want this for my PRs ASAP lol

uvcat7 (approved):
I don't see any other issues

poco0317:
will deal with cleaning up based on the reviews within a day or so

src/Core/Shader.cpp
poco0317: `log.clear()` would work as a replacement for `log = ""` but it would not allow `log.reserve(x)` to behave the same, so it would have to be changed to `log.resize(x)` unless `log.shrink_to_fit()` was also called.

The only use of these strings is for a log output and the reference is handed in by a function that hasnt even filled out the string. So it might just be better to not even reset the string here. But if future code uses these functions, the string might need to be reset. So I'll change it to `log.clear(); log.resize(size);`

src/Core/StringUtils.cpp
poco0317: It wasnt replaced. But now it is

sukibaby:
## Edit: can't test due to `Replace with FileReader/Writer with std::fstream` breaking building on my machine

<details>

Was anybody able to build and run this? if i build this PR on my PC, it compiles with a ton of warnings (shown in the build log below) but the program does not actually start.

[Output-Build.txt](https://github.com/user-attachments/files/21236879/Output-Build.txt)

i just get a blank window for a few seconds and then the program quits out

https://github.com/user-attachments/assets/799d6930-27d1-409b-82ae-dbf75beab6ca

</details>

nobbele:
It works fine for me

nobbele:
image

sukibaby:
Hiding my earlier comment as resolved,  re-cloning the branch fixed it, might have had some weirdness left over from when I was working on the cmake branch earlier.

Can anyone confirm if note tick is working normally with this PR or if it sounds like a weird electronic glitch sound?

sukibaby:
Squashed and merged. I had a lot of merge conflicts to resolve, so i might have gotten something mixed up along the way, or included something that was meant to be removed, so please feel free to look over https://github.com/uvcat7/ArrowVortex/commit/c969072d5e2e743cb41b001000cd176d2c52a8b9 and make a follow up PR if there is anything that needs changing.

poco0317:
the rebase you did undid a lot of the changes in #132 and i dont even know where to begin. @sukibaby @nobbele 
image
https://github.com/uvcat7/ArrowVortex/blame/c969072d5e2e743cb41b001000cd176d2c52a8b9/src/System/File.cpp

sukibaby:
@poco0317  all that FileReader stuff is dead code (completely unused) at the moment. PR #132 requires a followup, we wanted to handle that after merging these string changes - so they will be cleaned up then.


edit: we discussed it a bit more and will be reverting the squashed commit in beta and restoring the commits cherry picked from this commit as originally intended. the goal was to squash and merge so it could be easily bisected but that did not go as planned due to something in the git history.

poco0317:
you force pushed my clean commits out of that branch so it wont be very fun. might as well start over

StarbotArc:
Just letting you know, I am currently doing cherry-picks and re-merges on my side, you don't have to start over just yet.

sukibaby:
we kept your original untouched commits safe @poco0317 , at the moment @StarbotArc is handling whatever merge issues cropped up with the original PR and your original untouched commits will safely make it in.

poco0317:
my pr was perfectly fine until you rebased it. i had everything merged from beta already. you could have pressed one button.

sukibaby:
it actually got stuck on 23a9e8226daac2b4cee807d45e223f809f6d439a when attempting to do that

poco0317:
i dont believe that and it really really pisses me off a lot that i cant prove it because you've deleted the history by force pushing your rebase on this pr. all you had to do was say to fix a conflict if there was a conflict

StarbotArc:
Poco, please calm down. Look we definitely did screw it up, but we probably shouldn't be arguing about this. I'm trying to not waste your time by doing what I can right now to at least fix that issue.

poco0317:
my time is completely unwasted. you are wasting your own time

StarbotArc:
I know Poco, my time will forever be wasted, so I don't mind. I'd rather keep on going.

Issue #91, opened 2025-07-09
Replace File.cpp with std::fstream


uvcat7:

convextriangle:
closing since #132 #143 were merged

PR #132, opened 2025-07-13
Replace with FileReader/Writer and OpenFile with std::fstream

nobbele:
Addresses #91


nobbele:
The remaining functions in `File.{cpp,h}` are mostly to do with \<filesystem\> not \<fstream\> so I'm not migrating them in this PR. It'd be easier to migrate them after the string migration I think

uvcat7 (approved):
lgtm

PR #135, opened 2025-07-14
Migrate remaining Windows API calls in File.cpp to standard library

nobbele:


src/System/File.cpp
```diff
@@ -453,82 +399,60 @@ static bool HasValidExt(StringRef filename, const Vector<String>& filters)
...
 {
-	Vector<Path> out;
+	if (path.empty()) return {};
 
-	if(path.empty()) return out;
+	Vector<Path> out;
 
 	// Extract filters from the filter String.
 	Vector<String> filterlist;
-	if(filters)
```
DolphinChips, line 657: This entire chunk seems unrelated to the rest of PR. Reformatting of code would be nice, but belongs in another PR and we don't have code style guide for that unfortunately ):

nobbele: ugh auto-format

uvcat7: Eh, I think the autoformat changes are nbd. There's certainly not many of them.

nobbele:
non-ASCII paths broke in my earlier PR
I'd rather use UTF-8 `std::string`, but that's for later

src/System/File.cpp
sukibaby: Regarding a style guide, I started work for it to conform to the Google c++ style guide, but didn't get very far. I'd support new PR's following the Google style guide. I plan to continue with the style revision once the current wave of PR's is completed and merged.

sukibaby:
> non-ASCII paths broke in my earlier PR I'd rather use UTF-8 `std::string`, but that's for later

string changes are merged currently, if you'd like to update this accordingly

uvcat7:
Sorry for the merge/revert, I was hoping I could just quickly resolve the conflicts so I could send my review (I could not.)

PR #139, opened 2025-07-15
Correctly calculate frame counts when reading WAV files

nobbele:


sukibaby:
For future reference this is needed as a follow up to #130 .

sukibaby:
This was merged in with #130 .   @nobbele  was credited in the commit

PR #140, opened 2025-07-15
Revert "Replace String/StringRef with std::string (#130)"

sukibaby:
This reverts commit c969072d5e2e743cb41b001000cd176d2c52a8b9.

Follow up PR should choose cherry-pick commits of #130 into beta.


StarbotArc:
Cherry-picking for this instance will not work out, we may just have to fix it ourselves, or #130's author (poco0317) might make further patches.

PR #141, opened 2025-07-16
Fix a ton of warnings added after #130

poco0317:
So basically, #130 wasnt done


StarbotArc:
I'm guessing there's probably more in store you have to clean up?

uvcat7 (approved):
Seems good.

uvcat7:
The File.cpp changes reverted in #130 will be added back in #135, fyi.
I know there's still plenty of warnings even with these fixes, but a lot of them seem to have vanished in the CMake branch (#53), so I'm waiting to implement that before squashing the rest.

Issue #90, opened 2025-07-09
Upgrade project to C++ 20


uvcat7:

sukibaby:
If i try to build `main` as is as C++20 then I get 

- two instances of "can't convert Vortex::String to Vortex::String &"
- two instances of "can't convert const char array to Vortex::char_t *"
- `std::generate` in Polyfit.h needs replacing


I think if @StarbotArc ends up doing the refactoring of functions into their classes that may end up helping towards this cause.

And maybe just further replacing custom types/typedefs with standard types.

uvcat7:
Thanks @convextriangle for taking care of this!

PR #142, opened 2025-07-18
Resolve C++20 compilation

convextriangle:
Addresses #90;
By the way -- when testing, AV looks for `settings/settings.txt` and fails (because it isn't in the various source files?) - is this correct behavior?

### DIFF
--- a/build/VisualStudio/ArrowVortex.vcxproj
+++ b/build/VisualStudio/ArrowVortex.vcxproj
@@ -167,7 +167,7 @@
       <MultiProcessorCompilation>false</MultiProcessorCompilation>
       <ObjectFileName>$(IntDir)\dummy\dummy\%(RelativeDir)\</ObjectFileName>
       <AdditionalOptions>/FS %(AdditionalOptions)</AdditionalOptions>
-      <LanguageStandard>stdcpp17</LanguageStandard>
+      <LanguageStandard>stdcpp20</LanguageStandard>
     </ClCompile>
     <Link>
       <AdditionalDependencies>mad_debug.lib;libogg_debug.lib;libvorbis_debug.lib;libvorbisfile_debug.lib;freetype_debug.lib;winmm.lib;opengl32.lib;%(AdditionalDependencies)</AdditionalDependencies>
--- a/build/VisualStudio/ArrowVortex.vcxproj
+++ b/build/VisualStudio/ArrowVortex.vcxproj
@@ -137,7 +137,7 @@
       <MultiProcessorCompilation>false</MultiProcessorCompilation>
       <ObjectFileName>$(IntDir)\dummy\dummy\%(RelativeDir)\</ObjectFileName>
       <AdditionalOptions>/FS %(AdditionalOptions)</AdditionalOptions>
-      <LanguageStandard>stdcpp17</LanguageStandard>
+      <LanguageStandard>stdcpp20</LanguageStandard>
     </ClCompile>
     <Link>
       <AdditionalDependencies>mad_debug.lib;libogg_debug.lib;libvorbis_debug.lib;libvorbisfile_debug.lib;freetype_debug.lib;winmm.lib;opengl32.lib;%(AdditionalDependencies)</AdditionalDependencies>
@@ -200,7 +200,7 @@
       <MultiProcessorCompilation>true</MultiProcessorCompilation>
       <ObjectFileName>$(IntDir)\dummy\dummy\%(RelativeDir)\</ObjectFileName>
       <StringPooling>true</StringPooling>
-      <LanguageStandard>stdcpp17</LanguageStandard>
+      <LanguageStandard>stdcpp20</LanguageStandard>
     </ClCompile>
     <Link>
       <AdditionalDependencies>mad.lib;libogg.lib;libvorbis.lib;libvorbisfile.lib;freetype.lib;winmm.lib;opengl32.lib;%(AdditionalDependencies)</AdditionalDependencies>
@@ -234,7 +234,7 @@
       <MultiProcessorCompilation>true</MultiProcessorCompilation>
       <ObjectFileName>$(IntDir)\dummy\dummy\%(RelativeDir)\</ObjectFileName>
       <StringPooling>true</StringPooling>
-      <LanguageStandard>stdcpp17</LanguageStandard>
+      <LanguageStandard>stdcpp20</LanguageStandard>
     </ClCompile>
     <Link>
       <AdditionalDependencies>mad.lib;libogg.lib;libvorbis.lib;libvorbisfile.lib;freetype.lib;winmm.lib;opengl32.lib;%(AdditionalDependencies)</AdditionalDependencies>
@@ -268,7 +268,7 @@
       <MultiProcessorCompilation>true</MultiProcessorCompilation>
       <ObjectFileName>$(IntDir)\dummy\dummy\%(RelativeDir)\</ObjectFileName>
       <StringPooling>true</StringPooling>
-      <LanguageStandard>stdcpp17</LanguageStandard>
+      <LanguageStandard>stdcpp20</LanguageStandard>
     </ClCompile>
     <Link>
       <AdditionalDependencies>opengl32.lib;freetype.lib;mad.lib;winmm.lib;libogg_static.lib;libvorbis_static.lib;libvorbisfile_static.lib;%(AdditionalDependencies)</AdditionalDependencies>
@@ -302,7 +302,7 @@
       <MultiProcessorCompilation>true</MultiProcessorCompilation>
       <ObjectFileName>$(IntDir)\dummy\dummy\%(RelativeDir)\</ObjectFileName>
       <StringPooling>true</StringPooling>
-      <LanguageStandard>stdcpp17</LanguageStandard>
+      <LanguageStandard>stdcpp20</LanguageStandard>
     </ClCompile>
     <Link>
       <AdditionalDependencies>opengl32.lib;freetype.lib;mad.lib;winmm.lib;libogg_static.lib;libvorbis_static.lib;libvorbisfile_static.lib;%(AdditionalDependencies)</AdditionalDependencies>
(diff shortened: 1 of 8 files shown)


DolphinChips:
I also noticed that `settings/settings.txt` problem while testing CMake changes, I guess that's a regression from #132

nobbele:
there are some bad regressions there, I can look into it later

nobbele:
I'm not getting the settings.txt issue (beta + my PRs)
can you open a new issue with more details?

convextriangle:
turns out this is not related to the C++20 changes as far as I can tell

src/Editor/Music.cpp
```diff
@@ -94,10 +94,12 @@ MusicImpl()
 
 	bool success;
 
-	success = myBeatTick.sound.load("assets/sound beat tick.wav", false, std::string(), std::string());
+	std::string ignoredField;
```
uvcat7: This seems like a bad name for the variable. The field isn't ignored--it's just that no TITLE or ARTIST field is loaded. Something like emptyString would be better.

src/Editor/Aubio.cpp
```diff
@@ -81,7 +81,7 @@ void lvec_zeros(lvec_t *s) {
 
 /* MISC */
 
-uint_t fvec_set_window(fvec_t *win, char_t *window_type)
+uint_t fvec_set_window(fvec_t *win, const char_t *window_type)
 {
```
sukibaby, line 85: `char_t` is just a typedef for `char`.   Please first undo that typedef (and other Aubio typedefs) and then apply const here

src/Editor/FindOnsets.cpp
```diff
@@ -1043,7 +1043,7 @@ void FindOnsets(const float* samples, int samplerate, int numFrames, int numThre
 {
 	static const int windowlen = 256;
 	static const int bufsize = windowlen * 4;
-	static char* method = "complex";
+	static const char* method = "complex";
```
sukibaby, line 1046: this is passed to `new_aubio_onset`, please make sure types match

src/System/System.cpp
```diff
@@ -220,7 +220,7 @@ namespace {
 
 struct SystemImpl : public System {
 
-wchar_t* myClassName;
+const wchar_t* myClassName;
```
sukibaby: LPCWSTR would be more correct here.... this variable is to be used by `UnregisterClassW`, and then it shouldn't need the const qualifier

https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-unregisterclassw

src/Editor/Music.cpp
convextriangle: resolved in 682332ff0d92656e9958403e63bb7400c6dad0d3

sukibaby (commented):
I'd like to see more typedefs removed from the code base with this PR, at least the Aubio-specific ones. It also seems like we could benefit from changing some datatypes here (for example in `System.cpp` below).

src/System/System.cpp
convextriangle: resolved in e42b0b0b2150ac8c8dd932255d16212f4374851a

src/Editor/FindOnsets.cpp
convextriangle: https://github.com/convextriangle/ArrowVortex/blob/e42b0b0b2150ac8c8dd932255d16212f4374851a/src/Editor/FindOnsets.cpp#L957C1-L958C54

convextriangle: https://github.com/convextriangle/ArrowVortex/blob/e42b0b0b2150ac8c8dd932255d16212f4374851a/src/Editor/FindOnsets.cpp#L794

src/Editor/Aubio.cpp
convextriangle: for consideration / just asking - is addressing typedefs in-scope for a C++20 compilation hotfix PR? iirc blocking it for this might cause (ummm) more issues to fix for others who might add even more deprecated/removed/incompatible pre-C++20 features in their changes? this can be addressed in a different PR imo

sukibaby: > for consideration / just asking - is addressing typedefs in-scope for a C++20 compilation hotfix PR?

If it solves the underlying problem, I would say yes. But, I wrongly thought the typedef was causing an issue. The typedef is not the problem.  The problem seems to stem from the implicit conversion of strings elsewhere to const char*.

so for example these const's are not needed if explicit casts are added elsewhere.  for example:

image

since MSVC/VS2022 will typecast this to `const char*` without being cast, I guess the program is building with the `const` added to the function headers because the type safety rules enforce it.  but it seems like casting the arguments provided to `char_t*` to prevent them from being to `const char*` would maybe be the better option in terms of preserving original behavior.

convextriangle: personally i don't mind de facto constants being const - but maybe migrating to `aubio_window_type` for the window types would fit here (and using some sort of string lookups / etc. if we need the actual type string)?

DolphinChips: Given that aubio is a vendored C library, I'm in favor of moving vendored bits to a C source file and compiling it as such

sukibaby: > but maybe migrating to `aubio_window_type` for the window types would fit here

That seems reasonable.

Since I'm only seeing a `strcmp` comparison, I don't see why it doesn't just use the existing enum for the window type. If it turns out there is some reason not to use `aubio_window_type`, I can imagine a few ways to do this string comparison in ways that are more idiomatic for C++20 and up.

convextriangle: > Given that aubio is a vendored C library, I'm in favor of moving vendored bits to a C source file and compiling it as such

maybe doing this during/after cmake patching would be better? to avoid more breaking changes

src/Editor/Music.cpp
```diff
@@ -94,10 +94,12 @@ MusicImpl()
...
 	bool success;
 
-	success = myBeatTick.sound.load("assets/sound beat tick.wav", false, std::string(), std::string());
+	std::string emptyString;
+
+	success = myBeatTick.sound.load("assets/sound beat tick.wav", false, emptyString, emptyString);
 	if(!success) HudError("%s", "Failed to load beat tick.\n");
 
-	success = myNoteTick.sound.load("assets/sound note tick.wav", false, std::string(), std::string());
+	success = myNoteTick.sound.load("assets/sound note tick.wav", false, emptyString, emptyString);
```
sukibaby: Why do we have `emptyString` here? AFAICT it compiles just fine & everything is working when using  `std::string()` rather than `emptyString`.

Besides that PR looks good. 👍

convextriangle: > Why do we have `emptyString` here? AFAICT it compiles just fine when using `std::string()` rather than `emptyString`.

`bool Vortex::Sound::load(const char *,bool,std::string &,std::string &)': cannot convert argument 3 from 'std::string' to 'std::string &`
(the load function expects a few output references, creating the string in-place seems to break compilation on C++20)

sukibaby: Instead of one `emptyString` for both, prefer one for each field (e.g. `placeholderTitle` & `placeholderArtist`).

convextriangle: separated here 256dfd52f161b8c386a6b384eea70fc65168c0a7

build/VisualStudio/ArrowVortex.vcxproj
```diff
@@ -167,7 +167,7 @@
       <MultiProcessorCompilation>false</MultiProcessorCompilation>
       <ObjectFileName>$(IntDir)\dummy\dummy\%(RelativeDir)\</ObjectFileName>
       <AdditionalOptions>/FS %(AdditionalOptions)</AdditionalOptions>
-      <LanguageStandard>stdcpp17</LanguageStandard>
+      <LanguageStandard>stdcpp20</LanguageStandard>
```
sukibaby, line 170: One last thing.  there are some other `stdcpp17` in the vcxproj too.  For example, on line 237 the `stdcpp20` needs to be updated as well - possibly others


(this is why I failed to catch the `std::string()` failure: upon pulling the new commit, VS2022 switched back to C++17.)

sukibaby: @uvcat7   everything looks good to me once this is taken care of.

convextriangle: For Win32 / (possibly ARM) - these targets should probably be removed or handled somehow (only x64 would be left for now -- other vcxproj files don't target non-x64, and iirc [CMake configurations don't either](https://github.com/uvcat7/ArrowVortex/blob/92b44140e0b4b557f82b0853bfd27ac9b1444335/CMakeSettings.json#L1C1-L26C2))? x64 configurations should be working-ish
```
1>..\..\lib\libmad\obj\mad.lib : warning LNK4272: library machine type 'x64' conflicts with target machine type 'x86'
1>..\..\lib\libvorbis\obj\libogg.lib : warning LNK4272: library machine type 'x64' conflicts with target machine type 'x86'
1>..\..\lib\libvorbis\obj\libvorbis.lib : warning LNK4272: library machine type 'x64' conflicts with target machine type 'x86'
1>..\..\lib\libvorbis\obj\libvorbisfile.lib : warning LNK4272: library machine type 'x64' conflicts with target machine type 'x86'
1>..\..\lib\freetype\obj\freetype.lib : warning LNK4272: library machine type 'x64' conflicts with target machine type 'x86'
(and then ~27 linker errors)
```

DolphinChips: > [CMake configurations don't either](https://github.com/uvcat7/ArrowVortex/blob/92b44140e0b4b557f82b0853bfd27ac9b1444335/CMakeSettings.json#L1C1-L26C2))? 
>

This file is specific to Visual Studio, it might compile for aarch64 and others if one uses CMake directly

convextriangle: > This file is specific to Visual Studio, it might compile for aarch64 and others if one uses CMake directly

then likely current vcxproj compilation outside of x64 is irrelevant? it is not supported in the current beta branch and will likely be superseded in cmake changes

uvcat7 (approved):
lgtm

PR #157, opened 2025-07-31
wglSwapInterval set to 1 instead.

StarbotArc:
Setting ``wglSwapInterval`` to 1 because ArrowVortex doesn't need anything faster than the monitor's refresh rate.


uvcat7:
Setting the swap interval to -1 enables adaptive vsync. I changed this a few PRs ago.
Why would we not want to use that? I'm not understanding your explanation.

StarbotArc:
Adaptive vsync is broken in v1.0.1, and would take up a lot more power than necessary, which ideally an editor should not be doing.

uvcat7:
Okay, I finally figured out what is wrong here. I needed to check for `EXT_swap_control_tear` to enable adaptive vsync (I assume those for whom it was not working had graphics cards or OpenGL installs that didn't support it). I don't think having adaptive vsync vs. regular vsync is a big deal, but I was suspecting this was a different problem.

PR #168, opened 2025-08-10
Unique artifact names

ScottBrenner:
Appending SHA to the uploaded artifact's name, reference https://docs.github.com/en/actions/reference/workflows-and-actions/contexts#github-context

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -54,7 +54,7 @@ jobs:
         if: github.ref_name == 'beta'
         uses: actions/upload-artifact@v4
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/windows.yml
```diff
@@ -54,7 +54,7 @@ jobs:
         if: github.ref_name == 'beta'
         uses: actions/upload-artifact@v4
         with:
-          name: AV
+          name: AV-${{ github.sha }}
```
ScottBrenner:
```suggestion
          name: ArrowVortex-${{ github.sha }}
```

PR #171, opened 2025-08-11
Speed labels

uvcat7:
Fixes glitchy formatting errors with the 6-digit speed segment changes.


enc_temp_folder/2f8844bd56a22748fa2b9df87529a46/Segments.cpp
```diff
@@ -0,0 +1,732 @@
+#include <Simfile/SegmentGroup.h>
```
Psycast: Temp file include a mistake?

Psycast (approved):
Looks good to me.

PR #173, opened 2025-08-12
Circle Noteskin

Bamks:
Ported the "MultiplyByZero" noteskin from Etterna.


uvcat7 (approved):
Sure, looks fine. Why not?

PR #176, opened 2025-08-12
Google style guide formatting enforcing

### LINKED ISSUE #148: Unified formatting / tidying for the codebase
(sort of a nice to have) Likely would be useful to consider using tools like clang-format and clang-tidy to avoid common issues (and enforce those checks in CI)

convextriangle:


.clang-format-ignore
```diff
@@ -0,0 +1,2 @@
+lib/
+src/Core/ImageLoader.cpp
```
uvcat7: Why is this one file ignored?

convextriangle: heavy use of macros seems to break clang-format

uvcat7: k, add a comment explaining that to the format file and update ImageLoader.cpp to use the correct formatting by hand

convextriangle: removed the ignore with recent commits, no formatting errors in CI it seems

uvcat7 (approved):
Thanks a ton for this.

PR #179, opened 2025-08-12
MinGW build support on Linux

nobbele:
A lot is copied from `linux-4-mingw32` but I'm uploading it here as a reference (specifically the source code changes).
`settings.json` just has hardcoded values for my system.
I have not tested on Windows.


uvcat7:
Does #281 obsolete this? Or do you still want to do MinGW builds?

nobbele:
It obsoletes this

Issue #134, opened 2025-07-14
Not automatically switching to new chart


sukibaby:
How to reproduce:

- be on a chart already (Dance Single Challenge), press shift-N to bring up the "New Chart" dialog, set the difficulty (Dance Single Hard),  press Create.

Expected behavior per 2017 release:

- new chart is created, and editor switches to the new chart

Current behavior:

- new chart is created, but editor does not switch to the new chart

It seems like it takes a few clicks to select the other chart as well, at least off a fresh launch

PR #180, opened 2025-08-13
Sort charts on difficulty change + Open New Chart

### LINKED ISSUE #169: [Bug Report] Charts not always appearing in expected order in "List Of Charts"
**Describe the bug**
If you're rearranging difficulties of an existing simfile, they might not appear in the expected order until you close and re-open the simfile.

**To Reproduce**
Open a simfile with at least 1 existing difficulty.
Make a new chart.
Move that chart into an overlapping difficulty with the existing chart.
Move the old chart to a different difficulty.

**Expected behavior**
2017 build ArrowVortex re-sorts the list as expected, but AV 1.0.1 does not

**Screenshots**

Image

Psycast:
- Trigger a chart re-sort on chart difficulty change.
- Reset Chart Index on New Chart to allow switching for new charts.

When a new chart is added, all the charts are sorted and indexes change, but `openChart` will only switch if the index has changed since the same one, which won't happen if the new chart is sorted before the current chart.

Fixes #169, #134.

### DIFF
--- a/src/Managers/ChartMan.cpp
+++ b/src/Managers/ChartMan.cpp
@@ -135,6 +135,7 @@ struct ChartManImpl : public ChartMan {
 
             gSimfile->openChart(bound.chart);
             bound.chart->difficulty = newDiff;
+            gSimfile->sortCharts();
             gEditor->reportChanges(VCM_CHART_PROPERTIES_CHANGED);
         }
         return msg;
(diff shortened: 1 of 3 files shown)


uvcat7:
Can you redo this now that I merged in #176?

Psycast:
> Can you redo this now that I merged in #176?

Done.

PR #186, opened 2025-08-17
Simplify build system

DolphinChips:
Originally I tried to imitate what was original build system was doing aka building every folder in src as a static library. Turns out changing `src/System/System.cpp` with such setup and Ninja as a generator would cause rebuild of the *entire* project instead of just one file.


DolphinChips:
oops, should've removed draft a few hours ago

CMakeLists.txt
```diff
@@ -18,12 +18,25 @@ find_package(Ogg CONFIG REQUIRED)
 find_package(Vorbis REQUIRED)
 find_package(OpenGL REQUIRED)
 
-add_subdirectory(src/Core)
-add_subdirectory(src/Dialogs)
-add_subdirectory(src/Editor)
-add_subdirectory(src/Managers)
-add_subdirectory(src/Simfile)
-add_subdirectory(src/System)
+file(GLOB SRC "${CMAKE_CURRENT_SOURCE_DIR}/src/*/*.cpp")
```
convextriangle, line 21: just a thought - is the globbing approach better than explicitly listing files / the directory structure?
not sure but this can likely lead to some wacky build errors
e.g. [docs don't recommend this too](https://cmake.org/cmake/help/v4.1/command/file.html#glob)

uvcat7: We were globbing before, so I guess it's fine. I hate the idea of manually specifying the files; it seems unnecessarily complicated.
That being said it seems like having a CMakeLists.txt file in each directory is highly recommended.

convextriangle: as discussed, seems to impact build speeds
not so complicated since adding new files is not that hard?

uvcat7: Yeah, I talked to teejusb about best practices. We should do it "right" and manually specify every source file and stop using GLOBs.
Also when you create new files in VS it will update the CMakeLists.txt for you, so it'll be easy to do.

uvcat7:
Did this correctly in #222

PR #187, opened 2025-08-17
CD Title Preview in Simfile Properties

### LINKED ISSUE #158: [Feature Suggestion] a cd-title viewer below the banner would be nice
Examples

Image

Image

Image

uvcat7:
image

Adds a CD Title preview to the editor. The image is scaled to fit the available space if it is too large.

Closes #158.


LilSanik:
Is it compatible with animated cdtitles? like this 
cdtitle 8x1

uvcat7:
> Is it compatible with animated cdtitles? like this

No, but nothing in the editor is AFAIK. I'll add animation support as a future issue, but it'll be a lot more work than this, so don't expect it soon.

src/Managers/MetadataMan.cpp
```diff
@@ -214,7 +214,7 @@ struct MetadataManImpl : public MetadataMan {
                 return path.filename();
             }
         }
-        paths = File::findFiles(gSimfile->getDir() + "..\\", false);
```
convextriangle, line 217: just asking -- is it worth it to handle file paths manually without using [std::filesystem::path](https://en.cppreference.com/w/cpp/filesystem/path.html)?

convextriangle (approved):
test if performance things stay the same with large cdtitles on your machine too
e.g. i set the background of the chart as the cdtitle but it seems to work fine on my computer - likely a nitpick

uvcat7:
Looks like it's fine. Thanks!

PR #188, opened 2025-08-17
Clang errors

uvcat7:
I used the -fix argument on clang-tidy to squash the hundreds of warnings that our style guide was creating on build.
Think it's fine.

### DIFF
--- a/src/Core/ByteStream.cpp
+++ b/src/Core/ByteStream.cpp
@@ -11,7 +11,7 @@ namespace Vortex {
 // WriteStream.
 
 WriteStream::WriteStream() {
-    buffer_ = (uint8_t*)malloc(128);
+    buffer_ = static_cast<uint8_t*>(malloc(128));
     current_size_ = 0;
     capacity_ = 128;
     is_external_buffer_ = false;
@@ -19,7 +19,7 @@ WriteStream::WriteStream() {
 }
 
 WriteStream::WriteStream(void* out, int bytes) {
-    buffer_ = (uint8_t*)out;
+    buffer_ = static_cast<uint8_t*>(out);
     current_size_ = 0;
     capacity_ = bytes;
     is_external_buffer_ = true;
@@ -37,7 +37,7 @@ void WriteStream::write(const void* in, int bytes) {
         current_size_ = newSize;
     } else if (!is_external_buffer_) {
         capacity_ = max(capacity_ << 1, newSize);
-        buffer_ = (uint8_t*)realloc(buffer_, capacity_);
+        buffer_ = static_cast<uint8_t*>(realloc(buffer_, capacity_));
         memcpy(buffer_ + current_size_, in, bytes);
         current_size_ = newSize;
     } else {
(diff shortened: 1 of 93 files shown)


src/Simfile/LoadOsu.cpp
```diff
@@ -310,22 +310,22 @@ static void AssignDifficulties(Simfile* sim,
...
-                if (versions[i] == difficultyStrings[set][j]) {
+                if (versions[i] == difficultyString[j]) {
                     matches[j] = charts[i];
                 }
             }
         }
         int numMatches = 0;
-        for (int i = 0; i < 5; ++i) {
-            if (matches[i]) ++numMatches;
+        for (auto & matche : matches) {
```
Psycast: This should probably be `match`.

uvcat7:
I manually fixed another 200 or so warnings by hand. Pretty sure I got them all, but it'd be great if someone could verify.

src/Core/ByteStream.cpp
```diff
@@ -47,12 +47,12 @@ void WriteStream::write(const void* in, int bytes) {
 
 void WriteStream::write8(const void* val) {
```
convextriangle, line 48: fwiw it seems like writestream should be refactored out? in a future PR

src/Core/Canvas.cpp
```diff
@@ -303,21 +303,24 @@ void Canvas::box(float x1, float y1, float x2, float y2, float radius) {
...
     GetPolyDist func(x, y, vertexCount);
     data_->draw(canvas_data_, canvas_width_, canvas_height_,
-                {0, 0, (float)canvas_width_, (float)canvas_height_}, &func);
+                {0, 0, static_cast<float>(canvas_width_),
+                 static_cast<float>(canvas_height_)},
+                &func);
 }
 
 Texture Canvas::createTexture(bool mipmap) const {
-    uint8_t* dst =
```
convextriangle, line 317: not for this PR, future - addressing memory safety things?

src/Core/Slot.cpp
```diff
@@ -10,8 +10,8 @@ namespace Vortex {
 namespace {
 
 struct BaseValue {
```
convextriangle, line 12: for future too - Slot needs to be addressed?

convextriangle (changes requested):
no warnings on my machine
consider escalating warnings to errors in the clang-tidy config to force devs to check stuff like this

src/Core/Canvas.cpp
uvcat7: Yeah. There are a lot of mallocs/reallocs that don't check for error conditions.
I changed the ones that threw warnings to fail silently, but we should add exception handling and all that.

src/Core/Slot.cpp
uvcat7: This should be a bunch of static casts. Weird.

convextriangle:
> no warnings on my machine consider escalating warnings to errors in the clang-tidy config to force devs to check stuff like this

as in `.clang-tidy`:
```
WarningsAsErrors: '*'
```

also doublecheck if this does actually work in VS (e.g. do a C style cast somewhere and await the Armageddon)

uvcat7:
I'll add the clang-tidy enforcement in my next PR!

PR #191, opened 2025-08-20
Fix escaping for semicolon and pound characters

Exschwasion:
Seems like a typo that's causing `;` to be escaped as `\\#`, and `#` not being escaped at all.


uvcat7:
🤦‍♂️, thank you

PR #200, opened 2025-08-27
Integrate clang-format into the build

uvcat7:
* Adds clang-format error testing into the builds; like GitHub CI
* Adds a new target called "Reformat" that will fix clang-tidy and clang-format errors automatically (unfortunately, I could not find a non-janky way to do it)
* Fixed some style errors from my last PR
* Made the .rc file UTF-8 encoded so it will show Git diffs in the future (this actually threw an error during my testing)


convextriangle (approved):
works on VS2022
doublecheck for other platforms?

PR #205, opened 2025-09-03
Modify Shortcut Macro

Psycast:
Remove the comma so clang-format doesn't create that monstrosity.


uvcat7 (approved):
Thanks for fixing that.

Issue #198, opened 2025-08-23
[Feature Request] Convert to OGG from incompatible formats upon loading


sukibaby:
For my workflow, it's often necessary to convert audio to ogg, mp3 or wav, because I use the information in AV to decide where to make cuts in the audio.

I'd like to have the ability to convert incompatible forms of audio to ogg automatically. For example, FLAC audio or a webm file containing audio.

The interface I'd be imagining is:
1. Drop incompatible file on the AV window
2. Dialog appears saying this needs to be converted to ogg first, offer to convert to ogg.
3. Convert into a temporary directory if so.
4. Display some kind of progress information while the file is converted to ogg audio.
5. Newly converted ogg is loaded into the AV window.
6. If a simfile is saved with this converted audio, ask if we want to move the ogg into the folder the simfile is saved into as well

convextriangle:
side note - seems like an opportunity to integrate [ffmpeg](https://ffmpeg.org/)?

Psycast:
Would also like a export to MP3 option as well, due to game engine limitations only supporting mp3 on our side.

Issue #61, opened 2025-06-30
Build oggenc2.exe in the project


uvcat7:
See #15.

PR #215, opened 2025-10-03
Convert IIR1, stb_image, and aubio to vcpkg; replace oggenc2 with ffmpeg

### LINKED ISSUE #192: External library refactorings / updates
A lot of internal AV dependencies such as aubio or stb are outdated and may cause bugs for certain users (such as not being to load images in modern formats and such), all such instances should be indentified and either moved to lib/ or used via vcpkg.

uvcat7:
Closes #192 and #61.

* IIR1 (the Butterworth filter): This is the only change that could be a downgrade. Applying waveform filters is now a few times slower than it was before, but with the old C library it was quite snappy. I looked through both libraries and couldn't find the cause of the performance hit. If we want to actually fix this, we should use FFMPEG.

* aubio: This was only used for calculating onsets. Maybe the implementation is better now that it is updated; it's hard to tell. This also got rid of the FFT.cpp dependency.

* stb_image: I think the old copy of ImageLoader.cpp's stb_image was modified to add support for alpha-channel OpenGL textures. I had to manually hack that support back in.

* FFMPEG (oggenc2): I considered this a "bonus" dependency, so I suffered through setting up FFMPEG to convert files to .ogg. This change should also facilitate #198 with much less effort.

We should also consider using FFMPEG for loading audio files in the future, which should make it easy to support more audio formats.


Psycast (approved):
lgtm

PR #224, opened 2025-10-07
Freeform selection

DeltaEpsilon7787:
Better selection.

1. Can select both timing segments and notes while having a region defined all at the same time
2. Filtered selections like selecting only fakes, selecting only 8ths etc is now constrained by region
3. Minor cleanup of selection code in general with clang-tidy, though this one needs a full on intervention really.


src/Dialogs/AdjustSync.cpp
```diff
@@ -3,15 +3,18 @@
...
 #include <Core/WidgetsLayout.h>
 
-#include <Managers/TempoMan.h>
 #include <Managers/SimfileMan.h>
-#include <Simfile/SegmentGroup.h>
+#include <Managers/TempoMan.h>
 
-#include <Editor/View.h>
-#include <Editor/Selection.h>
+#include "Dialog.h"
```
uvcat7:
```suggestion
#include <Dialogs/Dialog.h>
```

uvcat7 (dismissed):
Looks good, other than the one nitpick.

PR #228, opened 2025-11-07
macOS Build Support

sukibaby:
Needed changes to allow building on macOS.

PREREQUISITES:

- brew install pkg-config
- brew install sdl3
- brew install nasm


uvcat7:
Looks fine so far. I would like to see some .yml files to prove the build process. Also not sure why it needs a build script; can you explain that to me?

Before I merge, I'd like to get the SDL porting work done so this could be used to iterate some Mac OS builds of ArrowVortex. Hopefully that won't take too long...

sukibaby:
It doesn't need the build script, sorry for the confusion. I'll clear that out.

And then hopefully I can get a yaml update here before this is merged as well :)

No worries on holding off on this for now. I don't expect much to change here.

sukibaby:
Re-opened and rebased following Linux support in beta. Working now on both x86_64 and arm64 macOS. Requires `pkg-config`, `nasm` and `sdl3` from Homebrew.

Screenshot 2026-09-01 at 10 07 11 PM

src/System/System.h
```diff
@@ -11,7 +11,10 @@ namespace fs = std::filesystem;
 #define MAX_PATH PATH_MAX
 #undef R_OK
 #endif
-#include <SDL3/SDL.h>
```
sukibaby, line 14: This `<SDL3/SDL.h>` include is actually duplicated in the file btw, so there is no issue with this line being removed.

uvcat7:
Thanks for doing this!
Can you add support for Mac builds to the .yml file, for both x64 and M1, and build information to BUILDING.md?

src/Core/Shader.h
```diff
@@ -2,6 +2,10 @@
 
 #include <Core/Core.h>
 
+#ifdef __APPLE__
+#include <OpenGL/gl.h>
```
uvcat7, line 6: Can you just include System/OpenGL.h instead to avoid the #ifdef spam? (and if so correct all instances of the include?)

src/System/OpenGL.h
```diff
@@ -1,14 +1,22 @@
 #pragma once
 
+#ifndef APIENTRY
+#define APIENTRY
```
uvcat7, line 4: Correct me if I'm wrong, but I thought Windows and SDL (Linux) both properly include the APIENTRY definitions. Maybe move these to the APPLE macro?

CMakeLists.txt
```diff
@@ -46,6 +49,19 @@ if("${toolchain_include}" STREQUAL "")
     endif()
 endif()
 
+if(APPLE)
```
uvcat7, line 52: Just wanted to say. This is cool.

src/System/OpenGL.h
sukibaby: Tbh I wasn't sure if this should be part of the Apple specific stuff or if we should just check if it's not defined. Np to move it to the Apple specific stuff.

 Also tbh I think we could just not build the debug stuff for macOS to prevent this from being an issue, but I felt that would be kinda jank.

uvcat7: Kinda weird that Apple's OpenGL headers don't define those macros for debug, if they don't yeah we should keep it.

PR #244, opened 2025-12-15
Feature goto quantize insert

robertpelloni:


uvcat7:
I don't accept low-quality slop in this repository, AI-generated or otherwise.

If you want to add new features to ArrowVortex, that's awesome. Buggy features that are impossible to to maintain are worse than no features at all. I can already see that several of these changes won't work as advertised without even running the code or reviewing it in detail.

Please do the following in the future when submitting PRs:
1. Submit against the beta branch. Some of these changes have already been implemented there.
2. Add or change one feature at a time. These changes are nearly impossible to review.
3. Explain what to me what each feature is supposed to do instead of submitting the code with zero comment or explanation. (Using the AI to do it for you does not count.)

If you don't put any effort into your submissions, you're ultimately wasting your time, my time, and the compute of whatever AI you used.

PR #248, opened 2025-12-24
Beatline row hover info

Psycast:
2 slightly different features that developed in tandem.

image

# Additional Beatline Options
- Use current snap
- Use snap colors
- Highlight mouse position.

Adds saving/loading of these options along with waveform & note display, which previously reset on close.

This changes the default shortcut name for Beatline toggle, breaking compatibility with older settings.txt, but makes it consistent with other setting groups.

Additionally snap colors used in minimap drawing have been moved to Common are are reused for the snap colors, with a slight tweaked blue color.

# Mouse Hover Position in Status Bar
Keep track of the nearest snapped row when hovering the notefield with the mouse.
A value of < 0 should be considered invalid.

Adds a option to display the row/beat/measure/time of the hovered row into the Status Bar.


uvcat7 (approved):
Excellent as always! These features are pretty handy.

PR #249, opened 2025-12-24
Clipboard, String Replace fixes.

Psycast:
- Make String replace no longer potentially infinite loop.
- Fix some code comments.
- Set default ClipboardData values.
- Only use notes within a region, not region itself, for minRow.


uvcat7 (approved):
lgtm!

Issue #220, opened 2025-10-03
[Feature Request]  Select By Number of Columns (Density)


LilSanik:
For Example, If I want to create a new difficulty using the jumps from the first diff as a base
By the number of coloumns. 1, 2, 3, or 4 (obviously, if the chart has more columns, the window will show more options)

<img width="667" height="231" alt="Image" src="https://github.com/user-attachments/assets/f17f84a8-9cda-4b8b-93d7-f6538e3b2843" />

<img width="584" height="592" alt="Image" src="https://github.com/user-attachments/assets/9aeaccce-26d3-4ccb-8ff0-faa4543160bd" />

Psycast:
Added in #250.

PR #250, opened 2026-01-01
Implement Select By Column Count/Density

Psycast:
This only select notes that match the density selected on a given row.

This ignores mines that would also be on the same row.

Key Shortcuts were added for 1-10, though the menu will display up to the SIM_MAX_COLUMNS of 32, matching the column count of the current sim file.

This fixes [#220](https://github.com/uvcat7/ArrowVortex/issues/220).


uvcat7 (approved):
Good as always.

uvcat7:
The minimap doesn't update after making a selection, but that's another issue from #224 that needs to be fixed separately.

PR #252, opened 2026-01-19
Add word wrapping to Tempoboxes.

Psycast:
image

Add wrapping to TextLayout and TempoBoxes, splitting when the next word would cross the width boundary.

Other benefit is TempoBoxes are less broken due to changing the overall font size in the settings file as the box height isn't fixed at 32px.


uvcat7 (approved):
Don't see any issues. Works as advertised!

PR #258, opened 2026-02-11
Various Fixes

### LINKED ISSUE #159: Adjust Sync sometimes fails, gives a detected BPM of "120.00 BPM :: 100%"
I'm using the release 1.0.1 build here.  I noticed this when trying to detect sync of a particular section starting from an arbitrary position with a 64th quantization. I selected the region, hit "Find BPM", and it gave me the 120BPM suggestion. This was unexpected to me, but I clicked "Apply BPM" anyway, and then clicked "Find BPM" again after that and it showed valid BPM suggestions as would be expected.

Image

After that point, I started recording the screen capture and attempted to reproduce what I just described above.

https://github.com/user-attachments/assets/f90ed1a7-4afb-4184-91cf-ba1356432739

There are some other bugs/minor annoyances demonstrated in this video as well, though I don't know if the mouse cursor being incorrectly tracked is ArrowVortex's fault or Windows' built in screen recorder causing that.

Psycast:
Each fix is a different commit for easier viewing.

**Report Note Selection for all actions.**
Updates the minimap in line with selections more accurately.
Adds Hud Msg when using Select All shortcut.

**Update NoteStats on Warp/Fake changes.**
This updates Chart Properties when changed.
Changes Note Density text to only factor real hittable notes, as warp based files or large fake usage will skew the value.

**Ignore default found BPM when none found.**
Inform the user no bpm was found instead of giving a bad result.
Fixes #159.

**Include notes at the exact end of region when clearing.**
This prevent left over notes when pasting notes, scaling notes, switching sides, and the stream generator.
Tempo region clearing already included this condition.

### DIFF
--- a/src/Dialogs/AdjustSync.cpp
+++ b/src/Dialogs/AdjustSync.cpp
@@ -162,7 +162,15 @@ void DialogAdjustSync::onTick() {
                 fmt.arg(i + 1).arg(t.bpm, 2, 2).arg(t.fitness * 100, 0, 0);
                 myBPMList->addItem(fmt);
             }
-            myApplyBPM->setEnabled(true);
+            if (myDetectionResults.size() == 0) {
+                myBPMList->addItem("- no results found -");
+                myBPMList->setEnabled(false);
+                mySelectedResult = -1;
+                myApplyBPM->setEnabled(false);
+            } else {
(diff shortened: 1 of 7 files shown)


uvcat7:
This looks great. Thanks for cleaning those issues up.

PR #266, opened 2026-03-07
File Saving Updates

### LINKED ISSUE #102: Make saving both .ssc and .sm the default save file format

Psycast:
Implements several quality of life features regarding saving and loading.

When using multiple default save formats, all are saved when using `Save`, but only the selected format is saved when using `Save As`.

**General:**
\- Add format feature flags to each segment type.
\- Add format feature count to quick test for format specific usage.
\- Auto-select .ssc / .sm based on SM5 feature usage on initial saving.
\- Add warning when saving a .sm when SM5 features are used.
\- Add multiple default save format support.

**OSU:**
Loading:
\- Prevent crash on blank Prop values.
\- Remove unused OSZ format.

Saving:
\- Fix saving from an audio file saving over the audio file.
\- Fix saving without any charts crashing.
\- Fix saving without any charts not actually saving.
\- Fix saving with charts not using the generated file names.
\- Fix saving with charts showing the wrong filename.

Fixes #102


uvcat7 (changes requested):
I don't see a setting for saving to just .ssc. Can you add one?

Also, you should add chart-specific tempo segments to the SM warning features list. I would also like the setting that saves both .ssc and .sm to only save .ssc if there is an .sm warning.

uvcat7:
~~Ok, that works, but there isn't an option to save both .sm and .ssc at the same time now. Can you add that?~~

NVM, the save formats are comma-separated. All good here

uvcat7 (approved):

uvcat7 (approved):

PR #267, opened 2026-03-22
Direct Tempo Editing

### LINKED ISSUE #60: Directly interact with notefield elements
Allow easy modification of BPM/stop/fake/snap/etc. values by clicking on them in the notefield.
That should pull up the dialog and let you change it.

This requires two features that don't exist yet:
-Dialogs need the ability to "grab" text input and select the existing text in a dialog box
-Selected dialogs should be keyboard shortcut compatible (i.e. Tab to move through display elements, arrow keys to modify dropdown values, etc.)

Psycast:
# Tempo Editing
Adds a mini editor for each tempo type.

Can be opened in 2 ways:
\- Using the dedicated hotkey for a given type, which will use either the mouse or receptor position based on the anchor setting.
\- Single selecting a tempo box.

The dialog cannot be moved by the user, and will be pinned to the location that it is currently editing, while staying on screen if scrolled out of view.

# Dialog Focusing
Adds a system when focus can be set directly with code, with any dialog opening attempting to focus on a widget marked as initial.

Currently unused, but functional.

# Other
\- Shortcut display now shows unassigned keys instead of not appearing in the list.
\- Dialogs can now have their Pin, Drag, and Close flags set for window buttons.
\- Dialogs can now be closed and reopened on the same tick.

Closes #60


uvcat7:
I like this, but the workflow is a bit clunky since you have to manually close the dialog box each time. Can you add support for the box closing when you hit Enter or Escape?

Psycast:
> I like this, but the workflow is a bit clunky since you have to manually close the dialog box each time. Can you add support for the box closing when you hit Enter or Escape?

Enter could be an issue due to some of the types having multiple inputs and the box only send the value changes on Enter or input loss, but I can check if Escape would lead to any issues.

uvcat7 (approved):

Issue #118, opened 2025-07-11
Don't save redundant BPM tempo markers


uvcat7:

uvcat7:
After some discussion, we decided it might be better to add a menu option to remove these, in case if they are wanted for some reason. It doesn't hurt anything to have them.

PR #268, opened 2026-03-22
Requantize / Recolor Notes

Psycast:
The allows for selected notes to be requantized / recolored to the current snap type, while preserving the bpm and scroll ratio.

This works be altering the bpm/scroll in a max 1 beat area around the note, with the bounds clamped to any notes or tempo that are closer to the note. If the area doesn't have a valid snap within the initial bounds, it will beat insert and try again.

Notes that are already the same snap type, or cannot be moved due to warps or other impossible

Known Issues:
\- Holds are converted to singles.
\- Beat Inserts will cause desync for files with multiple charts with shared timing.
\- Massive note batches can take a few seconds at worse.


uvcat7 (changes requested):
The requantization isn't selectable from the menu. 

Also, are all the extraneous tempo changes for this feature intentional? If not, can you fix that?

uvcat7:
My second request can be handled separately in #118.

uvcat7 (approved):

PR #269, opened 2026-03-22
Render when Resized / Menu Active

### LINKED ISSUE #221: [Bug Report] Editor area freezes during window resize
**Describe the bug**
If you resize an ArrowVortex window, the window will stop updating until the move is complete.

**Expected behavior**
The window should keep being redrawn and resized every frame.

Psycast:
This moves most of the render logic outside message loop so that it and an external timer can call render as needed.

Keeps track if the window is being resize or the menu is open, and if so allows a 60fps timer to trigger a render update as well.

The debug stats made this kinda ugly.

Fixes #221

### DIFF
--- a/src/System/System.cpp
+++ b/src/System/System.cpp
@@ -242,7 +242,7 @@ struct SystemImpl : public System {
 #ifndef NDEBUG
     long long frames = 0;
     int lowcounts = 0;
-    std::list<double> fpsList, sleepList, frameList, inputList, waitList;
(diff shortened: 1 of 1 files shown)


uvcat7:
Turns out, this issue is also fixed in #275. Since I'd need to revisit these code changes that PR, I'm just gonna close this one. Sorry we didn't coordinate here; I could have saved you some effort.

PR #270, opened 2026-03-22
Add reading MP3 metadata.

Psycast:
This is a feature supported when opening .ogg directly, so this adds mp3 support as well.

Supports most version of ID3.


uvcat7 (approved):
Looks great!

PR #272, opened 2026-03-28
Expand Audio Conversion

### LINKED ISSUE #162: [feature suggestion] .flac import
adding the ability to import .flac files would be good to reduce the time it takes for a map to get set up. 
Although this might not benefit games that support DDR-style and/or PIU-style panel layouts, this would be much better for rhythm games that only use arrow vortex for BPM timing and offset (e.g Beat Saber), since as of now you have to import the flac into audacity -> export the flac as ogg -> import the ogg into arrow vortex to get BPM and timing -> manually add the silence to the song with audacity -> export song as ogg again to import into the rhythm game.

### LINKED ISSUE #198: [Feature Request] Convert to OGG from incompatible formats upon loading

Psycast:
Add support for mp3 and wav conversion output.

Add a separate audio conversion that will attempt to convert any file it can find a valid audio stream in, basically supporting most things in ffmpeg.

Defaults filters for file selection are:
Audio: ogg, mp3, wav, flac, wma
Video: mp4, mkv, avi, mov, webm, flv

For multi-stream files, it will use the first audio stream found as the source.

Add menu options to convert simfile to ogg/mp3/wav, updating the simfile path to the converted audio.

Add file drop conversion for unsupported files that are convertible.

Fixes pts not being reset between conversions.

Fixes #198, #162


uvcat7 (approved):
This is awesome! Going to make audio conversions a lot easier for people.

PR #273, opened 2026-03-29
March Fixes

### LINKED ISSUE #264: [Bug Report] File Drop corrupts utf8 file paths.
**Describe the bug**
File Drop still uses old WideString and results in invalid/corrupted file paths result in loading failure.

**To Reproduce**
Drag a file or folder with utf8 characters in the path.

**Expected behavior**
The file should load like it does when using File > Load

Psycast:
Several Fixes:

FileDrop:
\- Fixes utf8 paths from not working, closes #264.
\- Fixes directory drops not finding any valid files due to extension now including the dot.

StringUtils:
\- Prevent empty string on readTime
\- Allow numbers starting with `.`
\- Prevent NaN as a result.

Dialog:
\- Clear request to open when already open.

Music:
\- Clear Artist on unload.

Noteskins:
\- Fix not finding a valid noteskin for a style if not in priority.
\- Fix not using the fallback noteskin if no valid noteskins are found.

CustomSnap:
\- Remove start capture text due to shortcut/input issues.


uvcat7 (approved):
lgtm

PR #276, opened 2026-04-24
Make the NPS Graph work more like Stepmania's

uvcat7:
As discussed in #251, these are my recommended changes to the density graph to make it work based on measure instead of time, to avoid holes or unexpected behavior.

This isn't a recreation of the Simply Love density graph--that is a bunch of quads drawn for each measure, which seems very inefficient for large files. Although it means that this graph looks bad at high densities since it is limited to a 1px maximum resolution.

To work around that, I added some sampling/averaging calculations to the NPS graph that should make the density visible for long simfiles without looking too weird.

The one thing this solution does not do is handle files that have very high BPMs for short times. Files should still work in that case, but it is possible the averaging will make the graph look unnecessarily blocky. That said, this should be pretty uncommon, so it's probably fine.

### DIFF
--- a/src/Dialogs/ChartProperties.cpp
+++ b/src/Dialogs/ChartProperties.cpp
@@ -267,9 +267,7 @@ class DialogChartProperties::GraphWidget : public GuiWidget {
    public:
     ~GraphWidget() override;
     explicit GraphWidget(GuiContext* gui);
-
     void updateGraph();
-
     void onDraw() override;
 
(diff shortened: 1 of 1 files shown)


Psycast (approved):
lgtm

uvcat7:
I reworked the algorithm to use time for rendering but measures for bucketing. Seems to work well for everything.

PR #278, opened 2026-05-21
Add DWI Saving Support +  Fix DWI Loading

SilentMystification:
### Summary

I have been making charts for [Dance Maniax Update](https://github.com/AllenSeitz/dance-maniax-update) which uses the DWI format. ArrowVortex is my preferred  step file editor and supports loading DWI files, but does not support saving DWI files.

Prior to this I had to go through an error prone multiprogram export to convert my ArrowVortex *.sm file to a *.dwi file. 

This PR aims to change that. There are 3 major things implemented here:

**Add DWI file saving support with full bracket coverage**
Implements SaveDwi to serialize ArrowVortex simfiles to the DWI format.

**Bug fix — LoadDwi {} quantization**
{ was incorrectly assigned quantization 4 (48th notes). The correct value per DMX and StepMania 3.9 is 3 (64th notes, 1/64 * BEATS_PER_MEASURE). Any DWI file containing {} brackets previously loaded with notes at wrong positions.

**Bug fix — LoadDwi closing bracket reset**
All closing brackets (), ], }, ') reset the quantization to 24 (8th notes, 24 rows/slot). The previous value of 32 has no correspondence to any standard note division in ArrowVortex's 192-row system (192 ÷ 32 = 6 slots/measure) and would place subsequent bare notes at wrong row positions. The correct reset value matches the parser's own initial quantization: 192 / 8 = 24.

### DWI Saving Details

**Metadata**
Writes #TITLE, #ARTIST, #GENRE, #FILE, #GAP, #BPM, #CHANGEBPM, #FREEZE, #SAMPLESTART, #SAMPLELENGTH, and #DISPLAYBPM. Timing offset is converted from AV's seconds-based offset to DWI's negated millisecond GAP. BPM changes and stops are written as comma-separated beat=value lists using DWI_beat = AV_row / 12.0.

**Charts**
Supports SINGLE, DOUBLE, COUPLE, and SOLO styles. Enforces DWI's one-chart-per-difficulty constraint with a warning on duplicates. Each chart header ends with a newline so note data starts on the following line.

**Note data — quantization**
Each measure selects the most compact bracket mode that covers all its notes:

Bracket | Grid | Rows/slot | Condition
-- | -- | -- | --
none | 8th | 24 | all rows divisible by 24
() | 16th | 12 | all rows divisible by 12
[] | 24th | 8 | all rows divisible by 8 (triplets)
{} | 64th | 3 | all rows divisible by 3
`' | 192nd | 1 | fallback

[] and {} cover disjoint note sets, so a measure mixing 24th and 64th notes falls back to 192nd.

**Note encoding**
Single note → numpad digit (4 2 8 6)
Two simultaneous notes → corner digit (7 1 9 3) or letter combo
3+ simultaneous → <ABC> multi-note bracket
Hold heads: X!X or PairChar!PairChar for 1–2; <ABC!ABC> for 3+
One tap + one hold head → CornerChar!HoldChar without brackets (e.g. 7!4 for Up tap, Left hold)


bin/settings/shortcuts.txt
```diff
@@ -1,6 +1,6 @@
 FILE_OPEN    = ctrl + o
 FILE_SAVE    = ctrl + s
-FILE_SAVE_AS =
+FILE_SAVE_AS = ctrl + shift + s
```
SilentMystification, line 3: I believe this makes sense as a default shortcut as many other programs use this exact combination to initiate the Save As flow.

src/Simfile/LoadDwi.cpp
```diff
@@ -199,12 +199,12 @@ static bool ParseNotes(Simfile* sim, char* p, int numPads, int numCols, const ch
...
-			case  '{': quantization = 4; ++n; break;
+			case  '{': quantization = 3; ++n; break;
 			case  '`': quantization = 1; ++n; break;
 			case  ')':
 			case  ']':
 			case  '}':
-			case '\'': quantization = 32; ++n; break;
+			case '\'': quantization = 24; ++n; break;
 			default:
 				n = ReadNoteRow(n, chart->notes, row, map, holds, quantization);
```
uvcat7, line 209: There's actually another bug here, though it would only have subtle effects when editing and copy/pasting: ReadNoteRow should be called with 192/quantization instead of quantization.

uvcat7 (changes requested):
Thank you so much for this pull request! I have a couple small things:

- Your edits/PR should have been made against the beta branch, which is the most up-to-date version of the code. This has caused some confusion in the past... sorry it bit you. It might be easiest to make a new PR.
- .dwi charts save from expert to beginner difficulty, when it should be the reverse.

Otherwise, this is very well-done!

uvcat7:
If you'd rather, I can make the new PR myself.

SilentMystification:
@uvcat7 

Apologies for the delays, i've rebased on beta and opened up a new PR: #294

PR #281, opened 2026-06-10
Port ArrowVortex to Linux

uvcat7:
Finally, we're ready for multiplatform support.
This PR was only tested on Windows WSL so far. I'll test more during the review process but this seems to work well enough for now.
The following changes were made:
- GCC did not handle Vortex::Vector properly and would throw runtime exceptions, so Vortex::Vector was ported to std::vector (this was a big change, as a lot of ArrowVortex's vector iteration did out-of-array access for position checking).
- min and max are part of MSVC so they were replaced with std::min/std::max/std::clamp as appropriate.
- ArrowVortex now behaves properly with Visual Studio 2026's clang-tidy (apparently 2022 was missing some files/checks?)
- Similarly, CMake Settings are deprecated in VS2026 and are not multiplatform, so CMakeSettings.json was converted to CMakePresets.json.
- I had so many problems with Visual Studio/CMake caching that I modified the CMakeLists.txt to work even when not called with the proper environment variables. In theory, this should make user presets easier to use, but still, it's CMake.
- Linux's clang-format gave some mutually exclusive results against Windows' clang-format, so the format checking runs only on Windows.
- A menu bar was written in OpenGL (Editor/Menubar.cpp, System/Menu.cpp) to behave the same as the Windows menu bar for Linux builds in lieu of messing with Wayland/X11. Note that to prevent rewriting all the rendering logic, the menu bar is rendered with negative coordinates and the rendering offset with glTranslated(). This can be revisited when SDL adds support for integrated menu bars.
image

- windows.yml has been renamed to release.yml and an entry for Linux builds on Ubuntu added.
- Of course, making C++ code that plays nicely with both GCC and MSVC deserves a special mention. Did you know that MSVC still doesn't implement std::aligned_alloc?

I would like some review of the CMakeLists.txt for any simplifications/optimizations.

For testing builds you may need to install all the packages listed in release.yml with your distro's package manager:
```
          sudo apt-get install -qq -y mono-devel nasm libltdl-dev \
            autoconf autoconf-archive automake libtool \
            libx11-dev libxft-dev libxext-dev \
            libwayland-dev libxkbcommon-dev libegl1-mesa-dev \
            libibus-1.0-dev zenity clang-tidy clang-format build-essential git \
            pkg-config cmake ninja-build gnome-desktop-testing libasound2-dev libpulse-dev \
            libaudio-dev libjack-dev libsndio-dev libusb-1.0-0-dev libx11-dev libxext-dev \
            libxrandr-dev libxcursor-dev libxfixes-dev libxi-dev libxss-dev libwayland-dev \
            libxkbcommon-dev libdrm-dev libgbm-dev libgl1-mesa-dev libgles2-mesa-dev \
            libegl1-mesa-dev libdbus-1-dev libibus-1.0-dev fcitx-libs-dev \
            libasound2-dev libdecor-0-dev
```


uvcat7:
I added build documentation for Linux as well.

Psycast (approved):

BUILDING.md
```diff
@@ -25,7 +25,63 @@ cmake --install build --prefix "<your_folder_here>"
...
+libdrm-devel mesa-libgbm-devel libusb1-devel libdecor-devel \
+pipewire-jack-audio-connection-kit-devel libthai-devel liburing-devel zlib-ng-compat-static 
+```
+
+You may be missing more packages depending on your distribution and what default packages it comes with. However, missing packages should be flagged by vcpkg or caused a build command to fail.
+
+### With CMake CLI
+In the terminal at the root folder of this project run:
+```
+cmake -Bbuild -S. -DCMAKE_BUILD_TYPE=Release -DCMAKE_TOOLCHAIN_FILE="$env:VCPKG_ROOT\scripts\buildsystems\vcpkg.cmake" -DVCPKG_TARGET_TRIPLET=x64-linux -DPRESET_NAME={linux-debug | linux-release}
```
convextriangle, line 66: this should probably be `$VCPKG_ROOT`, iirc `$env:<something>` is a powershell thing

BUILDING.md
```diff
@@ -25,7 +25,63 @@ cmake --install build --prefix "<your_folder_here>"
...
+Prerequsites:
+* Needed packages from package manager
+* CMake
+* vcpkg (per https://lindevs.com/install-vcpkg-on-ubuntu). Note you will need to give your user read/write access to the install folder with sudo.
+* `VCPKG_ROOT` environment variable pointing at vcpkg installation folder (env VCPKG_ROOT={path}) OR install it to /opt/vcpkg.
+
+### Build Packages
+For Ubuntu/Debian install:
+```
+sudo apt-get install build-essential git \
```
convextriangle, line 39: `python3-jinja2` might be missing on some debian installs

src/Core/Texture.h
convextriangle, line 1: `#include <vector>` seems to be missing in this header? seems to cause build errors at least on my machine

BUILDING.md
convextriangle: and `-Bbuild` seems to not be working; CMake presets seem to force `-Bout/build/linux-debug`? (same for the `cmake --build` command)

BUILDING.md
```diff
@@ -25,7 +25,63 @@ cmake --install build --prefix "<your_folder_here>"
...
+
+You may be missing more packages depending on your distribution and what default packages it comes with. However, missing packages should be flagged by vcpkg or caused a build command to fail.
+
+### With CMake CLI
+In the terminal at the root folder of this project run:
+```
+cmake -Bbuild -S. -DCMAKE_BUILD_TYPE=Release -DCMAKE_TOOLCHAIN_FILE="$env:VCPKG_ROOT\scripts\buildsystems\vcpkg.cmake" -DVCPKG_TARGET_TRIPLET=x64-linux -DPRESET_NAME={linux-debug | linux-release}
+cmake --build build
+```
+The resulting binary will be located at `out\install\{linux-debug | linux-release}\bin\ArrowVortex`
```
convextriangle, line 69: seems to be located in `out/build/linux-debug/src/System/ArrowVortex`, maybe this is for the installation, not building?

convextriangle (dismissed):

src/System/System.cpp
```diff
@@ -549,8 +520,6 @@ struct SystemImpl : public System {
                               SDL_WINDOWPOS_CENTERED);
     }
 
-    float getScaleFactor() const override { return myScale; }
-
     bool getWindowState() const override {
         auto temp = SDL_GetWindowFlags(window);
```
Psycast, line 524: Is the temp variable required?

BUILDING.md
sukibaby: I'm using 

`cmake -B build -S . -DCMAKE_BUILD_TYPE=Release -DCMAKE_TOOLCHAIN_FILE="$VCPKG_ROOT\scripts\buildsystems\vcpkg.cmake" -DVCPKG_TARGET_TRIPLET=x64-linux -DPRESET_NAME=linux-release -DCMAKE_C_COMPILER="gcc" -DCMAKE_CXX_COMPILER="g++" -DCMAKE_MAKE_PROGRAM="ninja"`

sukibaby:
I'm finally getting around to testing this on my Linux workstation... unfortunately I'm getting blocked atm by libsystemd failing to build its x64-linux-dbg build via ninja or make.

Will update if I make further progress here.

LetalexAlex:
@sukibaby Check out my automated compile script I made to compile the latest commit from the `linux` branch. https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=arrowvortex-linux-git

I had to patch vcpkg's libsystemd with `-Wno-error=override-init` on the meson file

My script although janky compiles on any ArchLinux machine i've tried with no errors, including clean roots.

Anyone feel free to use it as reference if you want.

sukibaby:
> @sukibaby Check out my automated compile script I made to compile the latest commit from the `linux` branch. https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=arrowvortex-linux-git
> 
> I had to patch vcpkg's libsystemd with `-Wno-error=override-init` on the meson file
> 
> My script although janky compiles on any ArchLinux machine i've tried with no errors, including clean roots.
> 
> Anyone feel free to use it as reference if you want.

ahhh, very nice, thanks. this does build fine.  :+1:  very much appreciated.

there is some room for optimization in the build, it takes 18m 50s even on a 18 core that usually crushes thru c++ lol.

given that this works, i'll compare it to the the version in this PR.


i think i've found a few bugs not present in the release 1.0.1 (pump noteskin not loading correctly by default for example) but i'll add an issue for anything i confirm is a bug vs the latest release. for now this is so great to have.  ty @uvcat7 !

sukibaby:
@uvcat7 can you try this and lmk if it still builds as expected on windows? splitting the sdl3 built via vcpkg to windows only seems to help, as it then uses the system sdl3 libraries on linux rather than needing to build them from source. i built on cachyos using system libraries this way and it works as expected.

03c2a618a218ba95ae6f57f82507b9e61e77866f

src/System/System.cpp
uvcat7: It just prevents hundreds/thousands of unnecessary calls to SDL_GetWindowDisplayScale, so I'll keep it
But I should just have the function scale an integer instead of having a bunch of static casts. I'll fix that.

src/Core/Texture.h
uvcat7: Sure enough, must be that the includes are masking the issue on my machine's GCC.

BUILDING.md
uvcat7: Something got wonky with the build output location. Should be fixed now.

uvcat7:
> @uvcat7 can you try this and lmk if it still builds as expected on windows? splitting the sdl3 built via vcpkg to windows only seems to help, as it then uses the system sdl3 libraries on linux rather than needing to build them from source. i built on cachyos using system libraries this way and it works as expected.
> 
> [03c2a61](https://github.com/uvcat7/ArrowVortex/commit/03c2a618a218ba95ae6f57f82507b9e61e77866f)

This doesn't work for my WSL installation, simply because libsdl3-dev is not available there... I guess I'll leave it the way it is.

src/Editor/Selection.cpp
```diff
@@ -280,8 +288,12 @@ struct SelectionImpl : public Selection {
             if (keyFlags & (Keyflag::SHIFT | Keyflag::ALT)) {
                 Texture& tex =
                     (keyFlags & Keyflag::SHIFT) ? myAddIcon : mySubIcon;
-                Draw::sprite(tex, {start.x, start.y, mpos.x - start.x,
-                                   mpos.y - start.y});
+                Draw::sprite(
+                    tex,
+                    {start.x - static_cast<int>(8 * gSystem->getScaleFactor()),
```
Psycast: This will need the gSystem->applyScaleFactor changes.

PR #284, opened 2026-06-11
Tweaking tweak

### LINKED ISSUE #160: [Feature Suggestion] Nested "Tweak BPM" windows
For the record, I have not tried using AV's new sync features. This is a different workflow which may or may not apply to the new sync tools.

Currently, if I click the wrench to tweak BPM once, then click it again on another point in the timeline, the previous BPM change will be cancelled.

My suggestion is to allow multiple tweak BPM contexts to coexist in a pile, editing the most recent wrench click and going backwards until the first one, possibly swapping between two if a previous timestamp is referenced.

Image

uvcat7:
Fixes #160 by having new tweaks apply the old one.

Also applies SDL scaling changes for the tweak boxes since it was missed.

### DIFF
--- a/src/Editor/Editor.cpp
+++ b/src/Editor/Editor.cpp
@@ -523,13 +523,14 @@ struct EditorImpl : public Editor, public InputHandler {
         if (gSimfile->isClosed()) return false;
 
         // Make a list of all simfiles in the current pack.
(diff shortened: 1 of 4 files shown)


Psycast (approved):

Psycast (dismissed):
Editor.cpp changes seem to be included from PR #283

PR #285, opened 2026-06-12
Fix mouse coordinates on HiDPI displays

LetalexAlex:
Fixes mouse coordinates being completely wrong on HiDPI displays on Linux, it seems to work correctly (at least for now) for me, please review the code before merging since I am not really the best programmer in the world

### DIFF
--- a/src/Core/TextLayout.cpp
+++ b/src/Core/TextLayout.cpp
@@ -187,8 +187,8 @@ static void ReadFontChange(const uint8_t* param, int len) {
 
 static void ReadFgColor(const uint8_t* param, int len) {
     if (LD->fgQuad.enabled) {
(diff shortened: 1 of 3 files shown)


LetalexAlex:
Mouse motion was converted to pixel space but button events were not, so clicks/hover didn’t line up. I used SDL_RenderCoordinatesFromWindow and SDL_GetWindowSizeInPixels so thge input now matches the OpenGL viewport

uvcat7:
Thanks for that catch! I was testing on a display without high DPI.
I cleaned up the code to remove some more jank I'd introduced and pushed it to the Linux branch; try it again and see if it still works for you. If so, you can close this PR.

LetalexAlex:
whoops I force-pushed to rollback on my side I was trying to fix widgets rescaling and moving things in strange ways when switching the window between my monitors (I have one set at 100% and another one at 135.something), what you made works absolutely fine on my side btw

PR #289, opened 2026-06-17
Add insert paste to edit submenu

DeltaEpsilon7787:
Just adding Ctrl-Shift-V to Edit submenu.


DeltaEpsilon7787:
what

DeltaEpsilon7787:
This makes no sense, why does it need conflict resolution?

PR #299, opened 2026-08-01
Visual Sync 2.0

DeltaEpsilon7787:
A rework of the original Visual Sync.

Features:
1. Draggable rows. Hold B or Alt-B to shift-drag rows around.
2. (Untested) support for split timed charts.
3. Visual Sync makes beatlines a hard requirement, they're now forcefully temporarily enabled whenever the user does anything related to visual sync, however they're ephemeral, they do not update user's settings and are disabled when the user does something else.
4. The notefield view will not move during syncing.

Note: due to shortcuts system skill issue, namely that it only supports button _presses_, the buttons for visual sync are fixed to B and Alt-B. This should likely be addressed at some point.
Note: Tempo changes no longer move charter's cursor along. While there might be an argument to be made for this behavior, I find it hard to justify.

Hopefully, with this the whole system also becomes as obvious as possible, no longer needing a whole separate page to figure it out.


uvcat7:
@DeltaEpsilon7787 can you remove the reordering of the header includes in the .cpp files? Just to keep the PR more targeted.
(It was throwing warnings for me, but it turns out that was because I was trying to build with VS2026 instead of VS2022.)

DeltaEpsilon7787:
I think the reordering of headers occurs due to autoformat, I did not specifically do it myself.

DeltaEpsilon7787:
Done.

uvcat7 (approved):
lgtm

PR #309, opened 2026-09-02
Update apt-get before calling to prevent Linux CI from failing on apt-get call

sukibaby:


uvcat7:
Thanks! Would be nice if this just worked (tm)

PR #315, opened 2026-09-02
Build song folder paths from UTF-8 instead of the ANSI code page

MAX-WiRED:
A song whose folder name is not spelled in the system code page loses its
artwork. The banner, background and CD title are never found, the fields in
File -> Properties stay empty, and stepping to the next song in the pack does
nothing. The chart itself opens, which makes it look like the images are
missing rather than the folder.

## Why

`SimfileMan::getDir` returns UTF-8 - it is filled from `pathToUtf8`. Several
places hand that string to the narrow `fs::path` constructor:

```cpp
fs::path sim_dir = fs::path(gSimfile->getDir().c_str());
```

On Windows that constructor decodes the bytes with the process code page. There
is no manifest asking for UTF-8, so it is whatever ANSI code page the system
runs - 1251, 1252, 932. A folder named `Плак-Плак` arrives as UTF-8 bytes, gets
read as if it were 1251 and turns into a different name that no directory has.

`findImageFile` then lists a directory that does not exist and returns nothing,
which is why the artwork silently disappears. `openNextSimfile` walks the parent
of the same broken path, and the relative path for a converted audio file is
computed against it too.

On Linux the narrow encoding is already UTF-8, so this only shows on Windows.

## The change

Use `utf8ToPath`, which `System/File.h` provides for exactly this and is already
used elsewhere in the same files:

```cpp
fs::path sim_dir = utf8ToPath(gSimfile->getDir());
```

Eight call sites across four files, no behaviour change for folders that happen
to be spelled in ascii.

## Reproducing

Put any song in a folder whose name is outside the system code page - Cyrillic,
Japanese, Greek - with a banner or background beside it. Open the chart and look
at File -> Properties: the image fields are empty and the previews are blank.
With the change the files are found and shown.

## Testing

Built and used in a downstream fork against a pack of songs with Cyrillic
folder names: banners and backgrounds are picked up, the previews render, and
moving to the next song in the pack works. Folders with ascii names behave as
before.

### DIFF
--- a/src/Dialogs/SongProperties.cpp
+++ b/src/Dialogs/SongProperties.cpp
@@ -339,7 +339,7 @@ void DialogSongProperties::myUpdateCdTitle() {
         auto meta = gSimfile->get();
         std::string filename = meta->cdTitle;
         if (filename.length()) {
-            fs::path path = fs::path(gSimfile->getDir().c_str());
+            fs::path path = utf8ToPath(gSimfile->getDir());
             path.append(stringToUtf8(filename));
             myCdTitleWidget->tex = extractSpriteSheet(path);
         }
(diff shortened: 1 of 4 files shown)


uvcat7 (approved):
Yep, that's a bug.

PR #318, opened 2026-09-03
Fix UTF-8 input for non-ASCII text

uvcat7:
Better fix for handling SDL's UTF8 input using standard C++ locale functions.
Supersedes #314.

### DIFF
--- a/src/System/System.cpp
+++ b/src/System/System.cpp
@@ -839,7 +839,8 @@ SDL_AppResult SDL_AppEvent(void* appstate, SDL_Event* event) {
         case SDL_EVENT_TEXT_INPUT: {
             const char* wp = event->text.text;
             const char n = '\n';
-            if (!std::iscntrl(*reinterpret_cast<const char8_t*>(wp), std::locale("")))
+            if (!std::iscntrl(*reinterpret_cast<const char8_t*>(wp),
+                              std::locale("")))
                 myEvents.addTextInput(wp);
             else if (*wp == '\r')
                 myEvents.addTextInput(&n);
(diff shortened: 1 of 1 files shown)


uvcat7:
@MAX-WiRED can you verify that this works for you on your non-US locales? This should also properly ignore other unicode format codes that should not be accepted as text input like DEL.

MAX-WiRED:
Tested on Windows 11, system locale `ru-RU`, ANSI code page 1251. Built this
branch and typed into the Subtitle field of Simfile properties, then saved and
read the bytes back out of the `.sm`.

Non-ascii input works:

```
#SUBTITLE:Привет こん é!;
```

which is `d0 9f d1 80 d0 b8 d0 b2 d0 b5 d1 82 20 e3 81 93 e3 82 93 20 c3 a9 21` -
Cyrillic, kana and latin-1 all intact.

DEL is gone, as you expected. Typing `c`, U+007F, `d` gives `cd` with no `7f`
byte in the file.

One thing worth knowing about the wider claim: only the **first byte** of the
sequence is tested, so this filters the ascii controls and nothing else. Typing
`a`, U+200B, `b`, U+00AD, `c` produces

```
#SUBTITLE:a<e2 80 8b>b<c2 ad>c;
```

Both go straight through, because their lead bytes are `e2` and `c2`, which are
not control characters. Filtering unicode format characters generally would
mean decoding the codepoint first, so if that is wanted it is a separate
change - the current one is not a regression either way, since those characters
were accepted before too.

Also, since `std::iscntrl` reads `LC_CTYPE` and ArrowVortex never calls
`setlocale`, the "C" locale is what is in force here, so the test is exactly
"not `00`-`1f` and not `7f`" no matter what the system locale is. That is why it
behaves the same on my machine as it would on a US one. Worth keeping in mind
if a `setlocale(LC_ALL, "")` ever shows up somewhere - on a single byte code
page some bytes above `7f` would start being classified differently.

Works for me either way. Thanks for picking it up.

PR #321, opened 2026-09-03
Add a frame rate limit and an FPS counter

MAX-WiRED:
How often the editor draws is not something the interface lets you touch. The
frame loop targets a constant:

```cpp
    // Non-vsync FPS max target
    auto frameTarget = duration<double>(1.0 / 960.0);
```

and vertical sync can only be turned off by editing `useVerticalSync` in
`settings.txt` by hand.

## The change

A new **Options** menu, between View and Help, holds **FPS limiter**:

```
Sync to display
────────────────
Limit to 60 FPS
Limit to 120 FPS
Limit to 240 FPS
Limit to 320 FPS
Limit to 480 FPS
Limit to 500 FPS
Limit to 960 FPS
Limit to 1200 FPS
────────────────
Unlimited
```

and **View → Show FPS** draws the current rate in the bottom left corner,
averaged over a quarter of a second. The counter is what makes the setting
worth having - without it there is no way to tell what the editor is doing.

The limit is one number: below zero waits for the display, zero runs free, and
above zero is a cap in frames per second. `System` keeps it in `sFpsLimit` and
`setFpsLimit` sets the swap interval to match.

Waiting for the display is a request a driver is free to ignore - on the
machine this was written on, "Sync to display" leaves the loop running at
several thousand frames - so the 960 ceiling stays underneath every mode except
"Unlimited". The default is therefore exactly what it is today: swap interval
1, capped at 960.

`disableVsync` is kept and now routes through `setFpsLimit`, so nothing that
calls it has to change.

## Settings

The chosen limit is written as `frameLimit`. The old `useVerticalSync` is still
written and still read, and is used as the starting value when `frameLimit` is
absent, so an existing settings file keeps behaving the way it did.

## Shortcuts

Every entry is an action of its own, so it can be bound to a key like anything
else: `SET_FPS_VSYNC`, `SET_FPS_60` … `SET_FPS_1200`, `SET_FPS_UNLIMITED`, and
`TOGGLE_SHOW_FPS`.

## Testing

The editor free-runs at about 200 frames on the machine used here. Picking
"Limit to 60 FPS" takes the counter to exactly 60, and the tick beside the
active entry follows the setting. The counter is drawn by the status bar, so it
appears once a chart is open.


uvcat7:
Convince me that we should have these features...!

There is already an FPS counter on the about page, so I don't know why we need another one.

Similarly I'm not sure what being able to adjust the FPS cap or remove it entirely helps the end user. AV is generally performant and I haven't run into any graphics issues (nor heard of people running into them) that would necessitate needing to change the cap from 960 FPS.

uvcat7:
As discussed in the Discord, I remain unconvinced that this will be a helpful feature.
I just don't think this will contribute anything to 99% of users. Thank you for suggesting it, though!

PR #323, opened 2026-09-03
Show the romanised names in the simfile properties

MAX-WiRED:
`#TITLETRANSLIT`, `#SUBTITLETRANSLIT` and `#ARTISTTRANSLIT` are read, written
and carried through every format, but there is nowhere to look at them or type
them. A song whose title is in Cyrillic or Japanese either arrives with
romanised names or it does not, and the editor cannot change that either way.

## The change

The simfile properties get the three fields, in a block of their own under the
names they belong to:

```
Title      Плак-Плак
Subtitle
Artist     IC3PEAK
Credit     WiRED
──────── Transliteration ────────
Title      Plak-Plak
Subtitle
Artist     IC3PEAK
```

The heading is there so the two `Title` fields are not mistaken for each other.

They behave like the fields above them: bound to the same metadata setters, so
a change is undoable and reaches the file on save.

## Testing

A song carrying `#TITLETRANSLIT:100sec. Kitchen Battle!!`,
`#SUBTITLETRANSLIT:pop'n music 8/Etterna Pack2` and
`#ARTISTTRANSLIT:Orange Lounge` opens with all three shown under the heading; a
song without them shows three empty fields.


uvcat7:
image

...This doesn't work if you uncheck the box after rechecking it, and there is blank space where the fields would go, so having the checkbox doesn't do anything useful.

I don't think there should be a checkbox to hide the fields at all. Why not make the properties box wider to make the transliterated fields go next to them, or put them directly underneath the corresponding non-transliterated field?

MAX-WiRED:
Dropped the checkbox — and with it the row-hiding code in **RowLayout** that it needed, so the bug you hit is gone along with what caused it. The romanised fields now sit in a block of their own below the originals, under a **Transliteration** heading so the two **Title** fields cannot be confused. I left the dialog width alone, since it is pinned in the side panel most of the time. Also rebased onto current beta

uvcat7 (approved):

PR #325, opened 2026-09-03
Export the song as an .osz archive

MAX-WiRED:
Handing a chart made here to osu! means packing it by hand: save the `.osu`,
find the audio and the background, zip the three together, and name the zip the
way the game names a download. **File → Export as .osz…** does that.

## What comes out

The same shape a downloaded archive has - I unpacked a few to check rather than
guess. A flat zip, no folders, every entry deflated:

```
IC3PEAK - Plak-Plak (WiRED) [Easy].osu
audio.ogg
bg.png
```

It holds what the file points at rather than whatever lies in the folder: the
chart files, the music and the background. A simfile may also carry a banner
and a CD title, but osu! knows neither, so they are left out.

The name is `Artist - Title`, using the romanised names where the simfile has
them, so a song called `Плак-Плак` by `IC3PEAK` is offered as
`IC3PEAK - Plak-Plak.osz`. The save dialog opens on that name and the
destination is the user's to choose.

## How

`SimfileMan::exportTo` writes the chart files into a directory of the caller's
choosing:

```cpp
    bool exportTo(const std::string& dir, SimFormat format) override {
        if (!mySimfile) return false;

        std::string songDir = mySimfile->dir;
        mySimfile->dir = dir;
        bool result = SaveSimfile(*mySimfile, format, false);
        mySimfile->dir = songDir;

        return result;
    }
```

The simfile keeps its own directory and its filename, and `onFileSaved` is not
called: an export is not a save, so the edit history still knows there are
unsaved changes if there were any. The exporter points it at a folder of the
program's own, packs from there and removes it, so nothing is left in the song
folder.

`ZipWriter` is the part of the zip format this needs and no more: one flat
level of deflated entries with their names stored as UTF-8. Entries are written
as they are added, so only one file is held in memory at a time. It uses zlib,
which the tree already pulls in through freetype - `vcpkg.json` now asks for it
by name rather than relying on that.

## Quaver

A `.qp` is the same archive with `.qua` files in it, so the same code covers it
in a few lines. That needs `SIM_QUA`, which is #316, so it is left out here
rather than tying the two together. Happy to add it once that lands.

## Testing

Exported a song with a Cyrillic title, romanised names, an ogg, a background,
a banner and a CD title. The archive passes an integrity check, holds the three
entries above and nothing else - the banner and the CD title stay out, and so
does a file in the folder that no tag names. Every entry is deflated, and
unpacking it into an osu! song folder gives a playable map.

The suggested name in the save dialog reads `IC3PEAK - Plak-Plak.osz`, and the
song folder is untouched afterwards - no stray `.osu` files.


sukibaby:
Not totally sold on this, I do see how it automates some of the preparation for osu but it seems like it adds a lot of complexity for a format that seems specific to osu. These kind of zips seem to be osu specific, and have a decent number of quirks to be aware of, so I worry that it adds a maintenance burden. Kinda prefer to keep AV's duties within managing the osu simfile itself.

PR #326, opened 2026-09-03
Export the song as a StepMania .zip, and ask which in a dialog

MAX-WiRED:
> Builds on the `.osz` export PR, which is where the exporter and the zip
> writer come from. The diff shown here includes that commit until it lands;
> the StepMania work is the second commit.

The archive a StepMania song travels in is not shaped like an osu! one. Taking
one apart shows why:

```
Unorthodox Red (debit13)/
Unorthodox Red (debit13)/DJ Nanasaki - Unorthodox Red.sm
Unorthodox Red (debit13)/DJ Nanasaki - Unorthodox Red.mp3
Unorthodox Red (debit13)/Unorthodox Red-bg.png
Unorthodox Red (debit13)/Unorthodox Red-bn.png
Unorthodox Red (debit13)/debit13.png
```

The song folder is inside the archive: unpack it into `Songs/<pack>/` and the
song is where the game expects it. An `.osz` is flat because osu! unpacks it
itself.

## The dialog

With more than one archive to choose from, the choice belongs somewhere it can
be seen, so the single `Export as .osz…` entry becomes `File → Export…` and a
window:

```
┌─ EXPORT ────────────────────────────────┐
│ Format   [ StepMania (.sm) in a .zip ▾] │
│ Name     [ Unorthodox Red (wired).zip ] │
│ [    Export...    ] [     Cancel      ] │
└─────────────────────────────────────────┘
```

| Format | Name |
|---|---|
| osu! beatmap (.osz) | `Artist - Title.osz` |
| StepMania (.sm) in a .zip | `<song folder>.zip` |
| StepMania 5 (.ssc) in a .zip | `<song folder>.zip` |

Picking a format rewrites the name, which is still free to edit, and the window
opens on the format the simfile came from. **Export…** then asks only where the
archive goes.

## What goes in

A StepMania archive holds one folder named after the song folder, with the
simfile, the music and the artwork inside it.

What goes in follows what the file points at, not what happens to lie in the
folder. StepMania reads a banner and a CD title as well as a background, so
those join the archive when the tags name them and the files are there; the
osu! archive still holds only the chart files, the music and the background,
which is all osu! knows about.

## ZipWriter

One addition: `addFolder`, a stored entry whose name carries a trailing slash,
with `0x10` in the external attributes so that a reader which ignores the slash
still sees a folder. `addFile` takes the folder as a prefix on the name, so
nothing else about the writer changes.

## Testing

A song folder named `Unorthodox Red (wired)` holding an ogg, a background, a
banner, a CD title and a file no tag names. The window opens on
`StepMania (.sm) in a .zip` offering `Unorthodox Red (wired).zip`; switching to
`osu! beatmap (.osz)` rewrites it to `IC3PEAK - Plak-Plak.osz` in front of you.

Exported all three ways:

```
Unorthodox Red (wired)/          Unorthodox Red (wired)/       IC3PEAK - Plak-Plak (WiRED) [Easy].osu
  .../song.sm                      .../song.ssc                audio.ogg
  .../audio.ogg                    .../audio.ogg               bg.png
  .../bg.png                       .../bg.png
  .../bn.png                       .../bn.png
  .../wrd_cdtitle.png              .../wrd_cdtitle.png
```

All three pass an integrity check, the unnamed file stays out of every one, and
the song folder is untouched afterwards.


sukibaby:
See https://github.com/uvcat7/ArrowVortex/pull/325#issuecomment-5545015552

PR #328, opened 2026-09-04
Fold the three anti-aliasing functions into one

MAX-WiRED:
The waveform's anti-aliasing is three functions that differ only in how many
samples they average:

```cpp
    void antiAlias2x(uint8_t* dst, int w, int h) { ... }
    void antiAlias3x(uint8_t* dst, int w, int h) { ... }
    void antiAlias4x(uint8_t* dst, int w, int h) { ... }
```

picked by a switch that has to grow an arm for every factor.

## The change

One loop covers every factor:

```cpp
    void antiAlias(uint8_t* dst, int w, int h, int factor) {
        const int newW = w / factor, newH = h / factor;
        const int area = factor * factor;
        ...
    }
```

and the switch becomes

```cpp
    if (getAntiAliasing() > 0) {
        antiAlias(texBuf, w, h, getAntiAliasing() + 1);
    }
```

The three named functions and the switch go. Nothing else about how a block is
rendered changes.

The list of settings is untouched: `None`, `2x`, `3x`, `4x`, as before. The
clamp that used to sit at the call site in `loadSettings` moved into
`setAntiAliasing`, so the bound is written once as `NUM_AA_MODES` rather than
as a bare `3`.

## Testing

Walked the setting from `None` to `4x` on a song with real audio: the list reads
as it did, the waveform redraws at each step, and `2x`, `3x` and `4x` look the
way they did before the change.

### DIFF
--- a/src/Editor/Waveform.cpp
+++ b/src/Editor/Waveform.cpp
@@ -28,6 +28,9 @@ static const int TEX_W = 1024;
 static const int TEX_H = 128;
 static const int UNUSED_BLOCK = -1;
 
+// None, 2x, 3x and 4x.
+static const int NUM_AA_MODES = 4;
+
 struct WaveBlock {
     int id;
     Texture tex[4];
@@ -185,7 +188,7 @@ struct WaveformImpl : public Waveform {
             if (ws) setWaveShape(ToWaveShape(ws));
 
             waveform->get("antiAliasing", &waveformAntiAliasingMode_);
-            setAntiAliasing(std::clamp(waveformAntiAliasingMode_, 0, 3));
+            setAntiAliasing(waveformAntiAliasingMode_);
         }
     }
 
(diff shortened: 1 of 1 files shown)


Psycast:
Personally anything past 4x is already past diminishing returns on quality, and is substantially more compute intensive.

uvcat7:
Refactoring the code to remove the three nearly-duplicate functions is worth it. But keep the max antialiasing at 4, please!

src/Editor/Waveform.cpp
```diff
@@ -289,10 +292,12 @@ struct WaveformImpl : public Waveform {
     WaveShape getWaveShape() override { return waveformShape_; }
 
     void setAntiAliasing(int level) override {
-        waveformAntiAliasingMode_ = level;
+        waveformAntiAliasingMode_ = std::clamp(level, 0, NUM_AA_MODES - 1);
         clearBlocks();
     }
 
+    int antiAliasingFactor() const { return waveformAntiAliasingMode_ + 1; }
```
uvcat7: Remove this function please and replace its calls with getAntiAliasing.

uvcat7 (changes requested):
Thanks! Just need to fix this one thing.

PR #329, opened 2026-09-04
Add one percent steps to the playback speed

MAX-WiRED:
Playback speed moves in tens:

```cpp
        CASE(SPEED_INCREASE)
        gMusic->setSpeed(gMusic->getSpeed() + 10);
```

which is a coarse step for the thing the setting is usually reached for -
slowing a hard passage down enough to read it, or nudging the speed until a
hand feels right. From 100% the next stop down is 90%, and there is nothing in
between, though `setSpeed` itself takes any value from 10 to 400.

## The change

A second pair of actions moves the speed by one percent, on Ctrl with the keys
the existing pair already uses:

```
Faster by 10%    Shift+Right
Slower by 10%    Shift+Left
Faster by 1%     Ctrl+Shift+Right
Slower by 1%     Ctrl+Shift+Left
```

`SPEED_INCREASE` and `SPEED_DECREASE` keep their keys and their ten percent, so
nothing anyone has learned changes. `SPEED_INCREASE_FINE` and
`SPEED_DECREASE_FINE` are new, and bindable like any other action.

The menu entries say which is which, so the difference is visible without
opening the shortcuts file.

## Testing

From 100%: `Audio → Speed → Faster by 10%` reports `Speed: 110%`, and
`Faster by 1%` then reports `Speed: 111%`. The submenu shows both pairs with
their keys - `Shift+Right` / `Shift+Left` on the ten percent step, `Ctrl+Shift`
on the one percent step - and `Default` still returns to 100%.


uvcat7:
Swap the shortcuts (so that shift is coarse and shift + ctrl is fine) so that you don't break everyone's muscle memory and that sounds good to me.

MAX-WiRED:
Swapped, thanks - **SPEED_INCREASE** and **SPEED_DECREASE** now keep both their keys and their 10%, so nothing anyone has learned changes. The 1% pair is new (**SPEED_INCREASE_FINE** / **SPEED_DECREASE_FINE**) and sits on Ctrl+Shift. That also made the patch purely additive.

PR #333, opened 2026-09-04
Add a Trim audio dialog for cutting a song down to size

MAX-WiRED:
Songs arrive with whatever sits in front of them - a few seconds of silence, an
intro nobody charts, a fade the recording came with. There is nothing here that
changes the audio itself, so the usual answer is to leave for Audacity and come
back, and then fix the offset by hand because the file no longer starts where
the chart thinks it does.

The idea came from a player who goes by **Sanik**, who asked for exactly this.

## The change

**Audio → Trim audio...** opens a dialog that cuts either end off the song,
fades it in and out, pads either end with silence, and moves the timing along
with the audio so the chart stays on the music.

```
First note at 2.575 s
[x] Trim the start
    Trim                    1.000
    [x] Measure from the first note
Last note at 99.063 s
[x] Trim the end
    Trim                    2.000
    [x] Measure from the last note
Fade in / Fade out          0.000 / 2.000
Silence at start / end      0.000 / 0.000
[x] Keep the original file
[x] Save the simfile afterwards
The wav is rewritten
Cuts 4.645 s
[Preview]  [Apply]  [Revert]
```

Cutting is the one thing here that throws audio away, so each end stays behind
a box of its own. With **Trim the start** and **Trim the end** both off, the
fields under them do nothing and only the fades and the silence apply.

**Measure from the first note** decides what the number means. On, it is the
silence to leave in front of the first note - the usual thing to want, and it
gives every song in a pack the same run-up whatever it started with. Off, it is
plainly how many seconds come off the front. Either way the cut is clamped so
the first note cannot end up before the song starts.

The end works the same way, which is what takes a stretch out of a long song to
be charted on its own. **Fade out** lands on the end the song is left with
rather than the one the file had, so a song cut mid-phrase eases out instead of
stopping dead. Cutting the end moves no timing: the offset only follows the
front.

The chart is not touched. Notes past a cut end stay where they are, over
silence - trimming is an operation on the audio, and throwing away notes is a
decision with its own undo entry.

Nothing touches the disk until **Apply**. **Preview** writes the result beside
the song as `<name> (trim preview).<ext>` and plays it, with the offset and the
song preview already shifted, so what you hear is what you would get. **Revert**
puts the song back and deletes that file. **Apply** either writes
`<name> (trimmed).<ext>` next to the original or takes its place, moving the old
file aside as `.old` the way a saved chart already does.

The song keeps the format it came in - trimming an mp3 does not quietly turn it
into an ogg. What that costs is on the dialog rather than in the manual: a line
reading `The mp3 is re-encoded` for a lossy song, `The wav is rewritten` for one
that is not, with the reason on the tooltip. Trimming decodes and encodes the
song again, and a lossy format gives up a little every time it is written.

## How it works

Both the cut and the silence happen in the FIFO between the decoder and the
encoder. The cut is drained off the front of it first, and the silence is put in
afterwards, so the silence lands in front of the music rather than being dropped
along with it. Frames on their way out are scaled by a gain while the song is
easing in or out.

Going through the FIFO rather than straight to the encoder is what keeps the
frames the size the encoder asked for: a stretch of silence rarely divides into
whole frames, and libvorbis refuses a short frame that is not the last one. The
silence after the song is the exception - nothing follows it but the flush, so
it is encoded directly and its short last frame is allowed.

The timing follows in one place, since the front loses the cut and gains the
padding: the first beat moves by the difference, and so does the song preview.

## Depends on #313

The Apply and Revert buttons disable themselves inside their own click handler,
and on `beta` a widget disabled while it holds the mouse never lets go, so the
dialog stops taking clicks after the first Apply or Revert - the buttons still
draw as usual and nothing happens when they are pressed. #313 is the one-line
fix; with it applied, Preview → Revert → Preview works over and over. Preview
itself is unaffected either way, because it is disabled from the tick after the
work is handed off rather than during the click.

## MP3 needs #337

Trimming an mp3 fails with `Failed to decode a frame into the FIFO buffer`.
That is not the trim: converting an mp3 through the existing **Convert simfile
to ogg** fails in the same place, and has all along - it leaves a nearly
complete file behind, so the failure is easy to miss. #337 fixes it, and with
that applied an mp3 trims like any other file.

## Testing

On an ogg song whose first note is at 1.643 s, with `#OFFSET:-0.552` and
`#SAMPLESTART:19.093`:

- Trimming to a 1.000 s run-up reports `Cuts 0.643 s`. After Preview the dialog
  reads `First note at 1.000 s` and plays `So_what (trim preview).ogg`.
- Revert deletes that file and puts the first note back at 1.643 s.
- Apply writes `So_what (trimmed).ogg` beside the untouched original and saves
  the chart with `#OFFSET:0.090909` (-0.552 + 0.643) and `#SAMPLESTART:18.450091`
  (19.093 - 0.643).
- One second of silence at the start with a 0.500 s fade in, and no cut, reports
  `Adds 1.000 s` and moves the first note to 2.643 s.

On a wav, where the arithmetic can be checked against the file itself - a song
whose first note is at 2.575 s, trimmed to a 2.000 s run-up: the 0.575 s cut
takes 101 356 bytes off a 16-bit stereo 44.1 kHz file, against 101 430 for the
exact length, the rest being the frame it rounds to. One second of leading
silence adds 176 400 bytes, which is one second to the byte.

Cutting the end, again on the wav where it can be counted: a song 104.133 s long
whose last note ends at 99.063 s, trimmed to leave 2.000 s after it, reports
`Cuts 3.070 s` and comes back 541 544 bytes shorter - 3.0700 s. With a 2.000 s
fade out the RMS over the last three seconds runs

```
1232 1007 1000 847 1641 1736 1393 982 472 253 128 53 21 5 1
```

per 0.2 s: flat until exactly two seconds from the new end, then down to
nothing. The same three seconds of the original file are silence, so the fade
had followed the cut rather than staying on the end the file used to have.

The line above the buttons reads `The mp3 is re-encoded`, `The ogg is
re-encoded` and `The wav is rewritten` on the three formats the editor loads.

With #337 applied, an mp3 whose first note is at 2.575 s trims to a 2.000 s
run-up and comes back as `100sec_Kitchen_Battle (trim preview).mp3` - still an
mp3, with `First note at 2.000 s`.

A failed trim leaves the folder exactly as it was.


sukibaby:
I would really want working support of all audio files formats AV supports to accept this, in other words this should support WAV and MP3 fully, as well as a notice that we're potentially re-encoding lossy audio if that is the case.

uvcat7:
I'd like this to support cutting the end of the song as well, with the option of adding a linear fadeout. This can be handy for cutting up parts of songs and making charts out them, which is a common action in our community.

MAX-WiRED:
> I'd like this to support cutting the end of the song as well, with the option of adding a linear fadeout. This can be handy for cutting up parts of songs and making charts out them, which is a common action in our community.

*Updated

Is that what you meant?

PR #340, opened 2026-09-08
Label memory corruption.

### LINKED ISSUE #334: [Bug Report] Adding multiple labels causes corruption and crash
**Describe the bug**
Using the Labels menu to add more than 2-3 labels will cause some labels to corrupt and display garbage data. Interacting with these labels in any way (attempting to delete, undo, edit, etc.) will crash ArrowVortex.

**To Reproduce**
1. Open any simfile.
2. Open the "Labels..." menu by going to Tempo > Labels...
3. Add a label with any text.
4. Scroll to another row, and add another label with any text.
5. Repeat steps 3 and 4 until the labels corrupt. Usually this happens after 2-4 labels are placed.

Attempting to undo, delete, cut, or edit the labels in any way will cause the program to crash

**Screenshots**

Image

**Additional context**
This bug occurs on the latest beta build (as of September 4 2026), built on Fedora 44 Linux.

Psycast:
Using short strings in Labels under 16 characters causes memory corruption under libstdc++ due to slight differences.

realloc/memmove does not update the internal pointers on short strings, MSVC avoids this by accident from using it's own address at retrieval time.

This replaces the old memory moving when inserting with std::move that handles std::string correctly.

(tbh I didn't test libstdc++, but it appears to work as intended, and can't be any worse then it currently is)

Fixes #334

### DIFF
--- a/src/Simfile/SegmentList.cpp
+++ b/src/Simfile/SegmentList.cpp
@@ -27,6 +27,21 @@ inline int diff(const Segment* begin, const Segment* end) {
            (reinterpret_cast<const uint8_t*>(begin));
 }
 
+static void relocate(const SegmentMeta* meta, void* dstPos, void* srcPos,
+                     int count, int stride) {
+    uint8_t* dst = static_cast<uint8_t*>(dstPos);
+    uint8_t* src = static_cast<uint8_t*>(srcPos);
+    if (dst > src) {
+        for (int i = count - 1; i >= 0; --i) {
+            meta->move(ofs(dst, i * stride), ofs(src, i * stride));
+        }
+    } else if (dst < src) {
+        for (int i = 0; i < count; ++i) {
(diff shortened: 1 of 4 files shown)


mangoafterdawn:
Now that I am able to use more labels, I discovered a crash that occurs while working in simfiles copy+pasting labels.  

1. Add a label, with any name below 16 characters
2. Copy the label to clipboard
3. Scroll to any other row, and paste the label
4. Create a new label on any row _above_ the pasted label. (Does not necessarily need to be above the initial label)
5. Ctrl+Z to undo the new label. Ctrl+Z once again to attempt to undo the pasted label, and observe the crash.

Sometimes, especially with more than just two labels, the pasted labels will have their text overwritten or partially corrupted by other labels. This bug does not occur with labels >=16 characters.

uvcat7 (approved):

uvcat7:
Note for myself that I should rewrite this to not play weird C-style memory games in the future.

